PR #23, opened 2026-06-03
Fix: URL-encode message IDs for Graph API path parameter

Davan-Etelamaki:
Message IDs from Microsoft Graph API use URL-safe base64 encoding with underscore and equals characters. These must be URL-encoded when passed as path parameters to avoid parsing errors.

### DIFF
--- a/src/outlook_mcp/tools/mail_read.py
+++ b/src/outlook_mcp/tools/mail_read.py
@@ -392,6 +392,11 @@ async def read_message(
     """
     message_id = validate_graph_id(message_id)
 
+    # URL-encode the message ID for the SDK path parameter
+    # Graph IDs use URL-safe base64 with _ and - which must be encoded
+    from urllib.parse import quote
+    encoded_message_id = quote(message_id, safe='')
+
     deferred_value: str | None = None
 
     if include_deferred_send:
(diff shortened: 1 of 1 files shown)


mpalermiti:
Thanks for this, and apologies it was closed without an explanation at the time.

The underlying instinct is correct — Graph IDs must be percent-encoded when they go into a raw path parameter — but this change would double-encode. `by_message_id()` is a Kiota SDK request builder, and Kiota already percent-encodes path parameters during URL-template expansion. Pre-encoding before handing the ID to the builder turns `%` into `%25` (and `=` into `%253D`), which Graph then 404s on. The same applies to the raw-URL branch in `read_message`, which already wraps the ID in `quote(message_id, safe='')` — the diff would `quote()` it a second time.

Encoding is in fact already handled everywhere an ID reaches a hand-built raw URL:

- `read_message` extended-property path — `quote(message_id, safe='')`
- `$batch` bulk read — `_build_read_subrequest_url()` → `quote(message_id, safe='')`
- delta queries — `mail_delta.py` → `quote(folder_id, safe='')`

…and the SDK builder paths encode themselves. So there's no unencoded-ID path left to fix.

Really appreciate you digging into this — please keep the PRs coming.

Davan-Etelamaki:
Took me a bit to have time to get back to this due to a project launch.  I actually closed the PR because I noticed the issue.

PR #28, opened 2026-08-23
Add uncategorized_only filter to list_inbox

jasond727:
## Why

Makes it easier for agents working on categorizing email to run a simple query to get the next batch of emails for processing. Previously there was no way to filter for messages with no categories assigned.

## What

- `outlook_list_inbox` (src/outlook_mcp/server.py): new optional `uncategorized_only: bool = False` parameter, passed through to the tool.
- `list_inbox` (src/outlook_mcp/tools/mail_read.py): when `uncategorized_only=True`, appends `not categories/any()` to the OData `$filter`, joined with any other active filters (unread_only, from_address, date range, classification) via `and`.
- Default is `False` — behavior is unchanged when the flag is omitted.
- New tests in tests/test_mail_read.py.

## Test plan

- [x] `uv run pytest tests/test_mail_read.py -k uncategorized -v` → 3/3 pass
  - uncategorized_only=True adds `not categories/any()` to the filter
  - default (flag omitted) adds no `$filter` at all
  - combined with unread_only + classification, all three clauses join via `and`
- [x] `uv run pytest tests/test_mail_read.py -q` → 27 pass
- [x] Full suite `uv run pytest --tb=no -q` → 548 pass, 13 pre-existing env failures (unrelated: calendar date + Windows symlink/permissions in test_config), 0 regressions from this change

### DIFF
--- a/src/outlook_mcp/server.py
+++ b/src/outlook_mcp/server.py
@@ -154,6 +154,7 @@ async def outlook_list_inbox(
     skip: int = 0,
     cursor: str | None = None,
     classification: str | None = None,
+    uncategorized_only: bool = False,
     concise: bool = False,
 ) -> dict:
     """List messages in one folder with structured filters (read, sender, date, Focused class).
@@ -165,6 +166,7 @@ async def outlook_list_inbox(
     Example: outlook_list_inbox(folder="Junk Email", unread_only=True, count=5)
     `folder` accepts display names, well-known names ("inbox", "junkemail"), or Graph IDs — prefer
     names. Pass concise=True to drop large fields (preview, categories) — ~10x fewer tokens.
+    Pass uncategorized_only=True to return only messages with no categories assigned.
     """
     client = _get_graph_client(ctx)
     return await mail_read.list_inbox(
@@ -178,6 +180,7 @@ async def outlook_list_inbox(
         skip,
         cursor=cursor,
         classification=classification,
+        uncategorized_only=uncategorized_only,
         concise=concise,
     )
 
--- a/src/outlook_mcp/server.py
+++ b/src/outlook_mcp/server.py
@@ -154,10 +154,10 @@ async def outlook_list_inbox(
     skip: int = 0,
     cursor: str | None = None,
     classification: str | None = None,
-    uncategorized_only: bool = False,
     concise: bool = False,
+    uncategorized_only: bool = False,
 ) -> dict:
-    """List messages in one folder with structured filters (read, sender, date, Focused class).
+    """List messages in one folder with structured filters (read, sender, date, category, Focused).
 
     Use this for folder-scoped browsing; use outlook_search_mail for KQL full-text search across
     all folders. For polling/recurring agents use outlook_list_inbox_delta (typically 10x cheaper
@@ -180,8 +180,8 @@ async def outlook_list_inbox(
         skip,
         cursor=cursor,
         classification=classification,
-        uncategorized_only=uncategorized_only,
         concise=concise,
+        uncategorized_only=uncategorized_only,
     )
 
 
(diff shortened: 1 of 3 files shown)


mpalermiti:
Thanks for this, and apologies for the slow response. I've approved the CI run so you can see it go green.

Before merging I verified the OData clause against a live personal outlook.com mailbox, since the unit tests can only assert that the string reaches `$filter`. It holds up: `not categories/any()` returns 200, composes with the existing `$orderby=receivedDateTime desc`, survives `$skip` and a full `@odata.nextLink` round-trip, and needs neither `ConsistencyLevel: eventual` nor `$count=true` (those are directory-object requirements, not mail). I also confirmed it actually discriminates rather than being silently ignored — created one categorized and one uncategorized draft, and only the uncategorized one came back. This is the first collection/lambda filter in the codebase, so that was worth checking.

Two small asks before I merge:

1. **Move `uncategorized_only` after `concise`** in both signatures rather than between `classification` and `concise`. `list_inbox` is importable from the published package and nothing after `graph_client` is keyword-only, so a mid-signature insert could rebind a positional `concise` for an external caller. No in-repo caller breaks — `server.py` passes keywords past `skip` — so this is purely defensive.
2. **Add "category" to the tool summary line** in `server.py`: `"...structured filters (read, sender, date, category, Focused class)"`. That first docstring line is what an LLM reads when choosing among 62 tools, so a new filter dimension should show up there.

Happy to make both myself on merge if you'd rather not push again — just say so.

One thing that is **not** yours to fix, flagging only so it doesn't surprise you: `list_inbox` sets `$orderby=receivedDateTime desc` unconditionally, which trips Graph's filter/sort compatibility rules. `from_address` and `classification` already return `400 InefficientFilter` on their own in the shipped release. So `uncategorized_only=True` combined with either of those will 400. That's pre-existing and I'm fixing it separately — I bisected every clause combination the function can build and your change adds no new failing ones.

mpalermiti:
Took the "happy to make both myself" option since this had been open a while — pushed the two changes to your branch rather than leaving it waiting on you. Merging now. Full credit stays on your commits; mine is a separate commit on top so it's clear what I touched.

What I changed:

1. Moved `uncategorized_only` after `concise` in both signatures.
2. Added "category" to the `outlook_list_inbox` summary line. Had to shorten "Focused class" to "Focused" to stay under the 100-col limit — `ruff` caught it at 105.

I also had to rebase you onto current `main`, which had one conflict. #32 landed in between and restructured exactly the block your change appends to: `list_inbox` now splits clauses into `date_filters` and `other_filters` so a `receivedDateTime` clause always leads the `$filter`. Your `not categories/any()` is a non-date clause, so it joins `other_filters` and picks up the floor. Nothing about your logic changed.

That's the fix for the `400 InefficientFilter` I flagged in my first comment — the one I said wasn't yours to fix. It's fixed now, and the combination that would have broken is verified working:

```
uncategorized_only=True                          200  n=3  (all uncategorized)
uncategorized_only=True, unread_only=True        200  n=1  (all uncategorized)
uncategorized_only=True, classification=focused  200  n=3  (all uncategorized)   <- was 400
uncategorized_only=True, after=2026-01-01        200  n=3  (all uncategorized)
uncategorized_only=True, concise=True            200  n=3  (all uncategorized)

CONTROL, pre-#32 clause shape:                   400   <- still fails, as expected
```

Read-only GETs against a live personal mailbox, filter strings captured from the actual code path rather than hand-written. Every returned message genuinely carried no categories, so the filter discriminates rather than being ignored.

577 tests passing, ruff clean.

Thanks for this — it's a genuinely useful primitive. Your PR description made the case for it well ("get the next batch of emails for processing"), and with mail-rules CRUD confirmed dead on personal accounts in #29, agent-side triage over `uncategorized_only` is now the supported path for programmatic inbox shaping rather than a convenience flag. Please do send more.

PR #29, opened 2026-08-25
Document messageRule as non-viable for personal accounts

tlewthwa:
Graph's docs for the `messageRule` endpoints (create/list/delete rule) list personal-account support via `MailboxSettings.ReadWrite`, but that's not borne out in practice — every operation returns `403 ErrorAccessDenied` on a real outlook.com/hotmail.com mailbox with that scope granted and consented. Verified with a raw HTTP probe against the live API directly (not just this project's client), so it isn't an SDK or auth-flow bug. Same class of doc/reality gap already documented for `/me/mailboxSettings` — just not flagged as unsupported in Microsoft's own reference pages this time.

This directly affects the "Mail rules CRUD" item under Near-term in ROADMAP.md, so I've flagged it there with a pointer to the full write-up under "Investigated and not viable", including a MAPI/COM (desktop Outlook automation) workaround that partially works — read/delete/stop-processing rules round-trip fine against the real web Rules UI — but hits its own separate wall on `MoveToFolder` specifically.

Docs-only change, no code touched.


mpalermiti:
Thanks for this, and apologies for the slow response. Genuinely appreciate that you probed the live API instead of taking the reference pages at face value — that's the half of this that's hard to get, and it's why the entry is worth having. CI run approved.

Two corrections before I merge, both in the framing rather than the finding itself:

1. **The contrast with mailbox settings no longer holds.** Both [`GET /me/mailboxSettings`](https://learn.microsoft.com/en-us/graph/api/user-get-mailboxsettings?view=graph-rest-1.0) and [`PATCH /me/mailboxSettings`](https://learn.microsoft.com/en-us/graph/api/user-update-mailboxsettings?view=graph-rest-1.0) now list personal-MSA support in their permissions tables — there's no "Not supported" row on either page today. So messageRules isn't a *different* class of gap from mailbox settings, it's the same one. Better to state the finding on its own terms ("docs list personal-MSA support; live API returns 403") and drop the comparison. This is as much on me as you: the existing mailbox-settings entry quotes doc language Microsoft has since changed, and I need to fix that line separately.

2. **The permissions quote is right for two of the three pages.** [Create rule](https://learn.microsoft.com/en-us/graph/api/mailfolder-post-messagerules?view=graph-rest-1.0) and [Delete messageRule](https://learn.microsoft.com/en-us/graph/api/messagerule-delete?view=graph-rest-1.0) do say `MailboxSettings.ReadWrite` for personal MSA, but [List rules](https://learn.microsoft.com/en-us/graph/api/mailfolder-list-messagerules?view=graph-rest-1.0) says `MailboxSettings.Read`. Small, but in an entry whose whole point is that these tables are unreliable, worth getting exactly right.

Would also be useful to state which scopes the probe token actually carried — that's the load-bearing detail for anyone re-testing this later.

One addition if you're up for it: `/me/inferenceClassification/overrides` works fine on personal accounts, and this repo already ships three tools over it (`outlook_set_inbox_override` and friends) for per-sender Focused/Other routing — no folder moves, but it's the one rule-shaped thing that does work. Since `errors.py` points every 403 at this exact section, a sentence saying "rules CRUD is blocked, but here's the narrow thing that works" would make the dead-end register actionable rather than purely negative.

The COM/MAPI investigation is a genuinely interesting result and I'm glad it's written down — no changes wanted there.

mpalermiti:
That's all three — thanks for turning it around so fast. Re-approved the CI run (the gate re-arms on each push from a first-time contributor, so that's on me to clear, not something you did wrong).

Checked the updated text against the live pages and it's accurate now: `MailboxSettings.ReadWrite` on Create and Delete, `MailboxSettings.Read` on List, each linked to the page it came from. Listing the probe token's scopes was the right call too — `MailboxSettings.ReadWrite` being present is what makes the 403 a real finding rather than a missing-scope artifact, and now nobody has to take that on faith to re-test it.

Two notes on how this lands:

**The mailbox-settings entry above yours turned out to be wrong**, and I've opened #33 to fix it. Your original framing pushed me to actually go read those pages, and the quote there — "documented as `Delegated (personal Microsoft account): Not supported`" — isn't on either mailboxSettings page and, going by the doc repo's history, wasn't there when 1.7.1 shipped in May either. It appears to be a misquote of a *neighbouring* `MailboxSettings`-scoped page (Get workHoursAndLocations, a different resource at `/me/settings/workHoursAndLocations`), which does carry that exact string. So your instinct that the two cases needed distinguishing was right — just in the opposite direction from how it was written. The empirical 403 on both stands.

I checked the merge both ways and there's no conflict, textual or semantic — your entry and the corrected one read consistently side by side.

**On the COM/MAPI paragraph:** during review someone raised that MS-OXORULE contradicts the "Delete is internally a special case of MoveToFolder" line, since OP_DELETE is its own ActionType. I dug in and that objection is wrong — OP_DELETE maps to `olRuleActionDeletePermanently`, not the `olRuleActionDelete` you're describing, and the spec is silent on how the latter serializes. Your sentence stands as written. Mentioning it only so it's on the record if it comes up again.

The one thing I'd still like, entirely optional and I'm happy to add it myself on merge: a sentence noting that `/me/inferenceClassification/overrides` works on personal accounts and this repo already ships three tools over it. Since `errors.py` sends every 403 to this section, it's the difference between a reader learning "this is impossible" and "this is impossible, but here's the narrow thing that works." Say the word and I'll append it, or push it yourself — either way this is ready.

Issue #52, opened 2026-09-12
Please enable private vulnerability reporting


neilbrencode:
@mpalermiti,  ```SECURITY.MD``` directs reporters to /security/advisories/new, but private vulnerability reporting is disabled on this repo, so that URL 404s for anyone without write access — there's no way to reach you privately.

Could you enable it under Settings → Advanced Security? I have something I'd like to report for your consideration with a tested fix ready, and I'd rather not describe it in public. Happy to open a PR the moment there's a private channel.

Thanks for sharing this repo,
Neil

mpalermiti:
Thanks Neil — you're right, and it's a real gap. SECURITY.md has been pointing at an advisory form that 404s for anyone without write access, so the policy was promising a private channel that didn't exist.

Private vulnerability reporting is now enabled — https://github.com/mpalermiti/outlook-mcp/security/advisories/new should work for you. Please go ahead, and open the PR whenever you're ready.

Worth flagging so you can judge whether your finding is already public: a security release is in progress covering findings the ClawHub scanner raised against this package, and that scanner report is public on the listing page. If yours overlaps, I'd still rather have your version — a tested fix from someone who actually looked beats a scanner line item. You'll be credited either way.

Thanks for handling it the careful way.

neilbrencode:
Thx!  It was the delta_token bearer disclosure issue that you've just fixed.

PR #54, opened 2026-09-12
fix: run on hosts without an IANA time zone database

### LINKED ISSUE #53: Windows: every calendar tool fails (no IANA tz database), and the test suite cannot run
Running outlook-mcp on Windows 11, every calendar tool fails, and the test suite cannot complete. All of it traces back to POSIX assumptions rather than anything Graph-related.

**Environment:** Windows 11 (10.0.26200), Python 3.12.13, uv 0.12.5, at `a6f6ac7`.

## 1. No IANA time zone database — every calendar tool fails

`outlook_list_events` returns a message-free failure:

```
Error executing tool outlook_list_events
```

The real cause, caught by calling the code path directly:

```
File "src/outlook_mcp/tools/calendar_read.py", line 29, in _compute_calendar_range
    tz = ZoneInfo(timezone)
zoneinfo._common.ZoneInfoNotFoundError: 'No time zone found with key America/Los_Angeles'

  ModuleNotFoundError: No module named 'tzdata'
```

`_compute_calendar_range` calls `ZoneInfo(config.timezone)` on every calendar read, and `validate_datetime` does the same for relative and zoneless dates. `zoneinfo` resolves those names against the host's IANA database, which Windows does not ship at all — and slim Linux images (`python:*-slim`, Alpine) routinely omit `/usr/share/zoneinfo`, so this is not Windows-only in practice.

`zoneinfo` already falls back to the `tzdata` package when there is no system database; nothing in `pyproject.toml` declares it. Installing `tzdata` fixes it completely, so the config value was valid all along.

## 2. The failure reaches the caller with no message

Worth separating from the cause, because it is what made this hard to diagnose: `ZoneInfoNotFoundError` is an *unexpected* exception to `_wrap_tool_errors`, which by design keeps that text server-side. The agent gets `Error executing tool outlook_list_events` and nothing else, for a condition with a one-line fix.

A missing database and a typo in `timezone` are also indistinguishable from the caller's side, despite needing completely different fixes. `validate_datetime` already routes the same config value through `ValueError` → `ToolInputError` so the message survives; the calendar path is the one that does not.

## 3. `time.tzset()` is POSIX-only — the suite cannot run

```
tests/test_validation.py:122: in test_naive_input_ignores_the_host_clock
    time.tzset()
E   AttributeError: module 'time' has no attribute 'tzset'
```

CPython only defines `time.tzset` on POSIX. The test dies before asserting anything, so the property goes unchecked on Windows entirely.

The assertion is really two claims under one name. That a zoneless string reads as `12:30Z` bare and `10:30Z` in `Europe/Berlin` needs no process timezone and can run everywhere; only the invariance claim — same answers whatever zone the host is in — needs `tzset`, and that part can skip on Windows without losing the reading.

## Suggested fix

I have items 1–3 fixed locally against `a6f6ac7`, with tests, `ruff` clean, and verified on the real Graph calendar path rather than only under mocks:

- add `tzdata` to `dependencies` — unconditional rather than `sys_platform == 'win32'`, since containers without `/usr/share/zoneinfo` hit the identical failure, and it is inert where a system database exists
- raise `ValueError` for an unresolvable zone, with separate messages for "no database" and "not a zone name"
- split the host-clock test so the reading is asserted everywhere and only the invariance check skips off POSIX

Happy to open a PR if that is welcome, and equally happy to adjust the dependency to a marker if you would rather keep it Windows-scoped. Let me know which you prefer.

Thanks for the project,
Neil

<sub>*- Claude Code*</sub>

neilbrencode:
Fixes #53.

On a host with no IANA time zone database, every calendar tool fails and the test suite cannot complete. Three commits, each standing on its own.

### `fix(deps): depend on tzdata so zoneinfo works off POSIX`

`_compute_calendar_range` calls `ZoneInfo(config.timezone)` on every calendar read, and `validate_datetime` does the same for relative and zoneless dates. `zoneinfo` resolves those against the host's database, which Windows does not ship and slim Linux images (`python:*-slim`, Alpine) routinely omit. `zoneinfo` already falls back to the `tzdata` package; nothing declared it.

Unconditional rather than `sys_platform == 'win32'`, because a container without `/usr/share/zoneinfo` hits the identical failure and a marker would leave it broken there. It is ~500KB and inert where a system database exists, since `zoneinfo` searches `TZPATH` first. Happy to switch it to a marker if you would rather keep it Windows-scoped.

### `fix(calendar): name the timezone that would not load`

`ZoneInfoNotFoundError` was an *unexpected* exception to `_wrap_tool_errors`, which deliberately keeps that text server-side — so the model received `Error executing tool outlook_list_events` with nothing in it, for a condition with a one-line fix. A missing database and a typo in `timezone` were also indistinguishable from the caller's side despite needing different fixes.

`_resolve_timezone` raises `ValueError` instead, the anticipated-failure channel `_wrap_tool_errors` converts to `ToolInputError`, carrying its message. This is what `validate_datetime` already does with the same config value. The two causes get separate messages, pointing at `config.json` and at the install respectively.

Also folds the duplicated `.astimezone(ZoneInfo("UTC")).strftime(...)` pair into one helper and uses `datetime.timezone.utc` — a fixed offset needs no database lookup.

### `test(validation): split the host-clock guard from the reading it asserts`

`time.tzset()` is POSIX-only, so `test_naive_input_ignores_the_host_clock` died with `AttributeError` before asserting anything.

Skipping it outright would have been cheaper and worse: it is two claims under one name. That a zoneless string reads as `12:30Z` bare and `10:30Z` in `Europe/Berlin` needs no process timezone and now runs everywhere. Only the invariance claim needs `tzset`, and that one skips off POSIX with a reason naming what still covers the reading. Both share `_both_readings` so they cannot drift apart.

## Verification

Windows 11 (10.0.26200), Python 3.12.13, uv 0.12.5.

- `ruff check src/ tests/` clean.
- `729 passed, 1 skipped, 4 failed`. The 4 failures are pre-existing on `main`, unrelated to these commits and untouched here.
- Exercised against the **live Graph calendar path**, not only the mocked suite: a real zone lists events, a typo'd zone returns the config message. The no-database branch is covered by patching `find_spec`.
- The POSIX-only test was run too, by standing in a no-op `time.tzset` before collection so the zone loop, the monkeypatch teardown and the assertion all really execute — 10 passed, nothing skipped.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


neilbrencode:
@mpalermiti Ready for your review

mpalermiti:
Thanks Neil — this is careful work and the diagnosis in #53 was better than what our own scanner had been telling us. The `ZoneInfoNotFoundError` point in particular: that's the same bug class we fixed in 1.20.0, surviving in a path the audit missed.

One blocker, then some small stuff. Heads up first: **I merged `main` into your branch** to resolve a `uv.lock` conflict — 1.21.0 shipped a wholesale dependency refresh for CVE fixes, and I regenerated the lock so `tzdata` lands on top of the refreshed pins instead of reverting them. Your three commits are untouched; you'll want to pull before doing anything else.

## Blocker: `_now_plus(0)` isn't the identity

`calendar_read.py:67-71`. PEP 495 — adding a `timedelta` to an aware datetime resets `fold` to 0. The old code derived `start` from `datetime.now(tz)` with no arithmetic, so it preserved `fold`. `_now_plus(0)` doesn't, and during the repeated hour after a DST fall-back the two disagree:

```python
tz = ZoneInfo("America/Los_Angeles")
# 09:30Z on 2026-11-01 = the SECOND pass through 01:30 local
now = datetime.fromtimestamp(datetime(2026,11,1,9,30,tzinfo=timezone.utc).timestamp(), tz)
now.fold                                    # 1
now.astimezone(UTC)                         # 09:30Z   <- old
(now + timedelta(days=0)).astimezone(UTC)   # 08:30Z   <- new
```

So `outlook_list_events(days=7)` with `after` omitted opens its window an hour early and returns events that already ended as upcoming. One hour a year per DST zone, only on the defaulted path — small, but it's a behaviour change in a refactor that reads as pure cleanup, and `test_days_widens_the_window_from_now` (`start < end`) can't see it.

Smallest equivalent fix:

```python
def _now_plus(days_ahead: int) -> str:
    moment = datetime.now(tz)
    if days_ahead:
        moment += timedelta(days=days_ahead)
    return moment.astimezone(dt_timezone.utc).strftime(_UTC_FORMAT)
```

That leaves `end` exactly as it was — including its own `fold` reset, which the original had too, so it's pre-existing rather than something to fix here.

**Worth flagging what *not* to do:** computing `start` in UTC (`datetime.now(dt_timezone.utc)`) fixes this case but isn't equivalent for `end`, because local-day arithmetic across a DST boundary is 23 or 25 hours, not 24. The window length would quietly change.

A regression test that freezes `now` inside the ambiguous hour and asserts `start == "2026-11-01T09:30:00Z"` would pin it.

## Smaller things, same push if you like

- **`_resolve_timezone` catches only `ZoneInfoNotFoundError`.** zoneinfo raises plain `ValueError` for path-shaped keys, so `/etc/localtime` — a plausible thing for someone to type — bypasses both your 50-char truncation and the config.json hint, surfacing zoneinfo's own text with the full key. `validate_datetime` already catches `(ZoneInfoNotFoundError, ValueError)`; matching it would close the gap.
- **`"Reinstall outlook-mcp"`** — the PyPI name is `outlook-graph-mcp`. Someone acting on that message searches for the wrong package.
- `find_spec("tzdata") is None` tests "tzdata isn't importable" rather than "there's no database". Probing `ZoneInfo("UTC")` inside the `except` would be exact. Low exposure now that tzdata is a hard dependency, so take it or leave it.

## Things I checked and liked

- The test split **increases** coverage rather than hiding the hard half: the invariance logic is identical to `main`, and keying the skip on `hasattr(time, "tzset")` rather than `sys.platform` means it still runs on every CI job.
- Unconditional `tzdata` is the right call, and your inertness claim holds — planting a fake zone on a private `TZPATH` confirms the system database wins and tzdata only fills keys it lacks.
- `ZoneInfo("UTC")` in the old code would itself have raised without a database. Your change removes a second failure point that neither of us mentioned.

Fix the first one and this is good to go.

neilbrencode:
Interesting - I had never heard of "fold", but it makes sense. I remember a number of Windows bugs when we fell back and hit 2am again :-)

neilbrencode:
Thanks — the `fold` catch is a good one, and you were right on all four points. I reproduced each before changing anything rather than taking them on trust. Pulled your merge first; 1.21.0's refreshed pins are in.

**The blocker.** Confirmed exactly as you described:

```
local            : 2026-11-01 01:30:00-08:00  fold = 1
old (no arith)   : 2026-11-01T09:30:00Z
_now_plus(0)     : 2026-11-01T08:30:00Z   fold after +0 = 0
```

Guarding the arithmetic restores the old value. I took your warning about computing `start` in UTC seriously and checked it: it fixes `start` and breaks `end`, since a local day across a transition is 23 or 25 hours.

The new test freezes `now` inside the repeated hour and pins the UTC value, plus the window length so a future "fix" can't quietly shorten it. I verified it **fails against the unguarded code** — two of its four assertions — and passes with the guard, because a regression test that passes either way isn't one.

Worth recording why nothing caught this: `start < end` stays true when `start` is an hour early, and comparing local datetimes would not have helped either, since PEP 495 has intra-zone comparison ignore `fold` as well — the two moments compare **equal** while converting to instants an hour apart. Only the UTC value pins it. No linter covers this; ruff's `DTZ` rules only flag naive construction. What would have caught it is a differential test asserting the refactor really is the identity: sweeping a year of hourly instants found exactly one broken hour in each of America/Los_Angeles, Europe/Berlin and Australia/Sydney, and none in UTC.

**The smaller three**, all taken:

- `_resolve_timezone` now catches `(ZoneInfoNotFoundError, ValueError)`. You were right that `/etc/localtime` escaped it — zoneinfo raises `ValueError("ZoneInfo keys may not be absolute paths")`, so it bypassed both the truncation and the hint. `../../etc/passwd` too.
- Install hint now says `outlook-graph-mcp`. `outlook-mcp` is the console script, not the installable — my mistake.
- Took the third rather than leaving it. `find_spec` tested importability, not whether a database exists; a POSIX host with `/usr/share/zoneinfo` and no `tzdata` installed would have been told its install was broken. `_has_time_zone_database()` resolves `ZoneInfo("UTC")`, which is the actual question.

Gate on Windows 11 / Python 3.12.13: `ruff` clean, **786 passed, 1 skipped, 4 failed** — the 4 being the pre-existing POSIX mode-bit tests, untouched here. Also re-ran the live Graph calendar path and both error paths.

<sub>*- Claude Code*</sub>

mpalermiti:
Merged — thank you, this was a genuinely good contribution.

The `fold` fix is right, and I checked it the way it deserved: reverted your guard and re-ran your new tests, and two of the four fail without it. That's a real regression test rather than one written to match whatever the code happened to do, which is rarer than it should be.

`TestAmbiguousHourKeepsItsFold` is the part I'll remember. `test_the_frozen_moment_really_is_ambiguous` guarding the premise is the kind of thing that keeps a test honest years later — if that instant ever stops being in the repeated hour, someone finds out instead of the real assertion quietly proving nothing. And documenting that the two datetimes compare *equal* because PEP 495 has intra-zone comparison ignore fold, so only the UTC value pins it, saves the next person the hour it would otherwise cost them.

You also took the three optional items, and `_has_time_zone_database()` is better than what I suggested — probing the capability instead of proxying it through `find_spec`, with the docstring saying why.

One trivial thing, already merged and not worth a PR: in `test_window_is_still_the_requested_length` the comment says *"7 local days across a fall-back is 7*24 + 1 hours"* but the assertion says `timedelta(days=7)`. The assertion is correct — both bounds land at 09:30Z. The comment describes the pre-fix behaviour, where `start` was misread as 08:30Z. Leftover reasoning, correct test.

For context on the branch churn: 1.21.0 went out while this was open, carrying a security fix and a wholesale dependency refresh, which is why I merged `main` into your branch rather than asking you to untangle the lockfile.

Your #52 is still open and the private advisory channel is live now, so whenever you're ready. Given how #53 and this turned out, I'm interested to see it.

PR #61, opened 2026-09-15
feat: real multi-account — per-capability routing, a cross-account gate, and aggregate reads

Nyaecho:
## Summary

The `accounts` config was scaffolding — `list_accounts`/`switch_account` existed, but nothing authenticated, stored, or routed per account: one credential served everything, whichever account it happened to be. This PR makes multi-account real, in three layers:

1. **Per-account auth** — each account gets its own token cache (`outlook-mcp-<name>`) and auth record (`auth_record-<name>.json`); `outlook-mcp auth <name>` authenticates it. The unnamed legacy path is untouched for single-account installs.
2. **Per-capability routing** (`capability_accounts`) — mail-centric toolset groups fold into `mail`, calendar/contacts/todo map 1:1, delta tools split by name, identity tools follow the active account. Routing is derived from the calling tool's name inside `_get_graph_client`, so the 60-odd tool bodies are unchanged and one `GraphServiceClient` is cached per account.
3. **Two independent switches** — `allow_cross_account` (deliberately switching routing; off = the agent sees one merged account, `switch_account` refuses *before* validating any name so refusals leak nothing, and no tool takes an account parameter) and `allow_aggregate` (three new fan-out read tools: `outlook_list_inbox_all` / `_events_all` / `_tasks_all`, concurrent per-account queries, account-tagged merged listings, per-account failure isolation into `errors`).

Also included: a guard on `outlook_changes_since` under split routing (it runs three deltas against one client and would silently mix accounts — it now refuses and points at the per-capability delta tools), and a full Chinese translation of the README (`README_CN.md` — happy to move it to a separate PR if you prefer).

## Repo conventions followed

- **CI guard updated**: `fresh-install`'s wheel-import assert 62 → 65 (`published-install` intentionally left at 62 — it checks the PyPI build, which is still 1.22.0).
- **RELEASING.md lockstep targets synced where they're feature-scoped**: `server.json` description count, `SKILL.md`/`README.md` counts and tables, `ROADMAP.md` (Ideas entry struck, Tier-1 "unused array" wording fixed), `CLAUDE.md` architecture listing (+ `routing.py`, `aggregation.py`). Version fields stay 1.22.0 for release time.
- **No new live-tier guards needed**: no `$filter`/`$orderby`/`$search` construction changed — routing/aggregate reuse the existing read calls verbatim.
- All four offline no-op guards pass (`no_dead_modules` / `no_dead_parameters` / `sdk_fields_exist` / `write_payloads_reach_the_wire`).
- Tool-surface budget: the 3 aggregate tools measured 12,043 → ceiling raised to 12,500 with the justification in the test file (the deliberate-raise path it documents).

## Testing

- 47 new offline tests (config validation, routing resolution, gate semantics, per-account storage naming, fan-out merge/sort/truncate/isolation, tool-name routing through the real mcp 2.x request context — including the dict-shaped `params` on `ServerRequestContext`).
- Adversarially verified on a real two-account setup: gate closed → schema audit shows none of the 65 tools exposes an account parameter, all switch attempts refuse (including to the default account and to guessed names), `list_accounts` collapses to one identity, routing is a pure function of config; gate open → per-capability and global switching work, configured routings still win over the active account. Aggregate reads verified merging real mailboxes (per-account tagged, globally sorted).
- Full suite: 818 passed; the 7 Windows failures are pre-existing on `main` (chmod/symlink semantics).

## 中文说明

`accounts` 此前只是配置摆设。本 PR 将其做实，分三层：**分账号认证**（独立令牌缓存与记录，`auth <账号名>` 分开登录，单账号安装完全不受影响）；**按能力路由**（`capability_accounts`，从调用工具名推导能力，60 余个工具签名零改动，按账号缓存客户端）；**两个独立开关**——`allow_cross_account` 总闸（关 = agent 只见一个合并账号，切换全拒且不泄露账号名；开 = 可全局或按能力切换，配置路由始终优先）；`allow_aggregate`（三个聚合只读工具，并发查询全部已认证账号，逐条带账号标签，单账号失败隔离进 `errors`）。

另含：分路由下 `outlook_changes_since` 守卫（防静默混账号数据）、README 中文全译。已按仓库规范同步 CI 断言（fresh-install 62→65）、server.json/SKILL/ROADMAP/CLAUDE 计数与条目；未改任何 `$filter`/`$orderby`/`$search` 构造，无需新增 live 层守卫。测试：47 个新离线测试 + 真实双账号对抗性验证 + 全量 818 过（7 个 Windows 预存失败与 main 一致）。无破坏性变更。


src/outlook_mcp/auth.py
```diff
@@ -89,21 +89,32 @@ def _is_azure_unencrypted_refusal(exc: BaseException) -> bool:
...
 
-def _auth_record_path() -> Path:
-    return Path(DEFAULT_CONFIG_DIR) / AUTH_RECORD_FILE
+def _cache_name(account: str | None) -> str:
+    """Keyring/DPAPI cache name — one per account.
+
+    The unnamed account keeps the legacy cache name so single-account installs
+    upgrade without re-authenticating.
+    """
+    return CACHE_NAME if account is None else f"{CACHE_NAME}-{account}"
```
mpalermiti, line 98: On macOS this does not give each account its own cache. azure-identity's `_get_persistence` builds `KeychainPersistence(<signal file per name>, "Microsoft.Developer.IdentityService", "MSALCache")` — the Keychain item is keyed by service+account only, so every `name` reads and writes the same item; the name only picks the signal file. `auth net` then `auth neko`: neko's cache starts empty (new signal file), its first save replaces the shared item with neko alone, and on the next `serve` net's silent refresh fails. Linux (libsecret keyed by name) and Windows (file per name) are isolated.

Fix: keep the one `CACHE_NAME` for every account and rely on the per-account `AuthenticationRecord` — MSAL caches are multi-account, and `authentication_record=` pins the identity. Then README's per-name cache claim and `cmd_logout`'s Keychain list go too.

src/outlook_mcp/auth.py
```diff
@@ -229,28 +267,69 @@ def _on_device_code(verification_uri: str, user_code: str, expires_on: object) -
...
+            fallback = next(iter(self._credentials))
+            logger.warning(
+                "Active account '%s' has no valid token; identity and unrouted "
+                "capabilities will be served by '%s' instead. Run "
+                "`outlook-mcp auth %s` to fix.",
+                self._active_account,
+                fallback,
+                self._active_account,
+            )
+            self._active_account = fallback
```
mpalermiti, line 310: This is fail-open. `resolve_capability_account` returns `_active_account` for every capability not in `capability_accounts`, so after this line every unrouted capability — writes included (`send_message`, `reply`, `forward`, `delete_message`, `batch_triage`, `create_event`) — runs on whichever account authenticated first, with `allow_cross_account=false`. `auth_status` says authenticated, `whoami` shows the other account, and the new INSTRUCTIONS line pre-explains that away. The docstring on `resolve_capability_account` ("pure function of the config") stops being true.

Please fail closed for data capabilities: leave `credential=None` so they raise `AuthRequiredError(default_account)` with the exact `outlook-mcp auth net` remedy — that is what routed capabilities already do. If identity tools should still answer, keep a separate identity-only fallback for `capability is None`. `test_try_cached_token_loads_each_account` currently pins the leaky behaviour as expected; flip it.

src/outlook_mcp/auth.py
```diff
@@ -260,10 +339,18 @@ def try_cached_token(self) -> bool:
...
             if _is_azure_unencrypted_refusal(exc):
                 raise UnencryptedTokenCacheError() from exc
-            logger.warning("Cached token refresh failed — re-run `outlook-mcp auth`.")
+            logger.warning(
+                "Cached token refresh failed for account '%s' — re-run `outlook-mcp auth %s`.",
+                account or "default",
+                account or "",
+            )
             return False
         except Exception:
```
mpalermiti, line 348: Combined with the fallback above, this makes the identity swap reachable by a transient failure: one 5xx or timeout from `login.microsoftonline.com` during startup on the default account, and the server runs as the other account for its whole lifetime.

src/outlook_mcp/auth.py
```diff
@@ -274,8 +361,63 @@ def get_credential(self) -> DeviceCodeCredential:
...
+    def get_account_credential(self, account: str | None) -> DeviceCodeCredential:
+        """Credential for an account by name; None means the legacy single account."""
+        if account is None:
+            return self.get_credential()
+        cred = self._credentials.get(account)
+        if cred is not None:
+            return cred
+        names = [a.name for a in self.config.accounts]
+        if account in names:
+            raise AuthRequiredError(account)
```
mpalermiti, line 373: Multi-account installs never see `startup_error` here. On Linux without libsecret, `lifespan` sets it (`UnencryptedTokenCacheError`) and `_credentials` stays empty, so every routed tool tells the agent `Run outlook-mcp auth net` — which fails identically. The single-account path (`get_credential`) surfaces the real remedy. Check `self.startup_error` first, as `get_credential` does.

src/outlook_mcp/auth.py
```diff
@@ -287,34 +429,54 @@ def list_accounts(self) -> list[dict]:
...
+                f"Unknown capability '{capability}'. "
+                f"Valid capabilities: {sorted(ROUTING_CAPABILITIES)}"
+            )
+        self._routing_overrides[capability] = name
+        return {"status": "switched", "capability": capability, "account": name}
+
+    def logout(self, account: str | None = None) -> dict[str, str]:
+        """Clear in-memory credentials and one account's auth record."""
+        if self.config.accounts and account is None:
+            account = self.config.default_account
```
mpalermiti, line 471: Two things. No-arg `logout` in a multi-account config narrows to the default account and unlinks only that record, while `cmd_logout` prints every account's cache name and says the server will require re-authentication — the other accounts' records survive and the next `serve` promotes one of them to active. And the next line nulls `self.credential` unconditionally, so `logout("neko")` while `net` is active leaves `net` in `_credentials` but `is_authenticated()` False.

Also worth a CHANGELOG line: `outlook-mcp logout` deleting the auth record is a behaviour change (main only printed instructions).

src/outlook_mcp/config.py
```diff
@@ -68,6 +122,52 @@ def _validate_allow_categories(cls, value: list[str]) -> list[str]:
...
+        - default_account / capability values must exist.
+        - default_account defaults to the first configured account.
+        """
+        if not self.accounts:
+            if self.capability_accounts:
+                raise ValueError(
+                    "capability_accounts is set but 'accounts' is empty — define "
+                    "the accounts first."
+                )
+            if self.default_account is not None:
```
mpalermiti, line 141: These validators reject config files `main` accepted: `default_account` set with `accounts` empty, or a name outside the new regex. `load_config()` is uncaught in `lifespan` (`server.py:59`) and in every CLI command, so an upgrade turns a working install into a server that dies at startup with the reason on stderr — and the operator cannot even `logout`. Either accept-and-warn for the legacy shapes, or catch in `lifespan`/CLI and print the fix.

Separately, `_ACCOUNT_NAME_RE.match` with `$` accepts `"net\n"` (`$` matches before a trailing newline); use `re.fullmatch`.

src/outlook_mcp/server.py
```diff
@@ -124,26 +136,92 @@ def _get_config(ctx: Context):
...
+
+    Multi-account configs route capabilities (mail / calendar / contacts /
+    todo) to accounts: the calling tool's name decides the capability, the
+    AuthManager decides the account, and one GraphServiceClient is cached per
+    account in the lifespan context. Single-account installs route everything
+    to the one credential, exactly as before.
     """
     auth = _get_auth(ctx)
-    credential = auth.get_credential()  # raises AuthRequiredError if unauthenticated
+    account = auth.resolve_capability_account(capability_for(_calling_tool_name(ctx)))
```
mpalermiti, line 174: Fails open: if `_calling_tool_name` returns `None`, `capability_for(None)` is `None` — the same value identity tools use — and the call routes to the active account, writes included, with no error or log. It works on mcp 2.2.0 today (`ServerRunner._on_request` puts the raw params mapping on `ServerRequestContext.params`), but that field is marked `TODO(L54): remove for Context rework`, `mcp[cli]>=2` is unbounded, `_calling_tool_name` accepts only a plain `dict` though the SDK types it as `Mapping` (middleware may hand back another mapping), and the tests here use hand-built `MagicMock` contexts.

Please: accept `collections.abc.Mapping`, raise when accounts are configured and a non-account-group tool resolves to no capability, and add one test through the real SDK (`Client(mcp)`) asserting a mail tool and a todo tool reach different credentials.

src/outlook_mcp/server.py
```diff
@@ -1071,6 +1158,90 @@ async def outlook_delete_task(
...
+    concise: bool = False,
+) -> dict:
+    """Inbox across every authenticated account, newest first, each message tagged `account`.
+
+    Requires allow_aggregate=true. One account failing lands in `errors`;
+    unauthenticated ones in `skipped_unauthenticated`. No cross-account cursor —
+    deep pagination is what outlook_list_inbox is for.
+    """
+    config = _get_config(ctx)
+    aggregation.require_aggregate(config)
```
mpalermiti, line 1183: These three tools are registered for every install and refuse unless `allow_aggregate` is set — the default. Measured with the surface budget's own recipe they cost ~464 proxy tokens per turn (3.9% of the surface), which is why the ceiling had to move. I don't have a demand signal for them yet (no `switch_account` calls in a month of my own traces), so I'd rather hold them for a separate PR. If they come back: prune them in `lifespan` when the gate is off (`toolsets.configure` already prunes), and give the refusal a real code and `action` like `ReadOnlyError`, rather than a plain `ValueError` that reaches the model as `invalid_input`.

src/outlook_mcp/aggregation.py
```diff
@@ -0,0 +1,229 @@
...
+    carried (OutlookMCPError embeds one in str()).
+    """
+
+    async def _one(
+        account: str | None, client: Any
+    ) -> tuple[str | None, list[dict] | None, str | None]:
+        try:
+            return account, await runner(client), None
+        except Exception as exc:  # noqa: BLE001 — the whole point is isolation
+            return account, None, str(exc)[:_ERROR_TEXT_CAP]
```
mpalermiti, line 72: The blanket `except Exception` turns caller-fixable input errors into a *successful* empty listing — `outlook_list_tasks_all(status="bogus")` returns `{tasks: [], errors: [<same message per account>]}` where `outlook_list_tasks` raises `ToolInputError` — and stringifies raw Graph SDK errors without `wrap_graph_error`, so a 401 loses `run outlook-mcp auth`, a 429 loses the back-off hint, a 404 loses `re-list`. Let `OutlookMCPError`/`ValueError` propagate (they're the same for every account), and run the rest through `wrap_graph_error` before stringifying.

src/outlook_mcp/aggregation.py
```diff
@@ -0,0 +1,229 @@
...
+            entry["account"] = account or "default"
+            tagged.append(entry)
+    tagged.sort(key=sort_key, reverse=reverse)
+    return {
+        items_key: tagged[:count],
+        "count": min(len(tagged), count),
+        "accounts_queried": accounts,
+        "skipped_unauthenticated": skipped,
+        "errors": errors,
+        "has_more": len(tagged) > count,
```
mpalermiti, line 114: This is only the truncation signal; each runner discards the per-account `has_more`, so the merged flag is `False` whenever the accounts' items fit in `count` — one account with 100 messages and `count=25` returns 25 with `has_more: false`. The module docstring promises the opposite ("whether any account had more"). Either OR in the per-account flags, or change the docstring and the field.

src/outlook_mcp/aggregation.py
```diff
@@ -0,0 +1,229 @@
...
+    docstring's no-cross-account-pagination stance.
+    """
+    tagged: list[dict] = []
+    for account, entries in per_account.items():
+        for entry in entries:
+            entry["account"] = account or "default"
+            tagged.append(entry)
+    tagged.sort(key=sort_key, reverse=reverse)
+    return {
+        items_key: tagged[:count],
```
mpalermiti, line 109: The per-account helpers clamp `count` to 1..100; this slice uses the raw value. `count=0` fans out to every account and returns `[]` with `has_more: true`; `count=-1` silently drops one item and reports `count: -1`. Clamp once in the three server wrappers.

mpalermiti (changes requested):
Thank you for this — it's a serious piece of work. The config validation and the gate design are right, and it exposed that the `accounts` scaffolding on `main` was broken (nothing ever filled `_credentials`, so `switch_account` blanked the credential). I want the auth/routing core. I can't merge this shape, for two hard reasons and one scope reason. Requesting changes; specifics are inline.

**1. The per-account token cache does not exist on macOS.** azure-identity's `TokenCachePersistenceOptions.name` only picks the *signal file* there; the Keychain item is always (`Microsoft.Developer.IdentityService`, `MSALCache`) — see `azure/identity/_persistent_cache.py:_get_persistence`. So `auth net` then `auth neko`: neko's cache starts empty and its first save replaces the shared Keychain item with neko alone; on the next `serve`, net's silent refresh fails and every tool routed to net raises `AuthRequiredError('net')`. Linux (libsecret keyed by name) and Windows (file per name) are fine; macOS is where most people run this. Fix: keep ONE cache name and rely on the per-account `AuthenticationRecord` you already save — MSAL caches are multi-account by design, and `DeviceCodeCredential(authentication_record=…)` pins the identity. That also removes the per-name Keychain entries from README and `logout`.

**2. The startup fallback is fail-open identity substitution.** When the default account's token fails at startup, `_active_account` is re-pointed to whichever account authenticated first, and every capability *not* in `capability_accounts` — reads and writes: `send_message`, `reply`, `forward`, `delete_message`, `create_event` — then runs on that account, with `allow_cross_account=false`. `auth_status` says authenticated, `whoami` shows the other account, and the new INSTRUCTIONS line tells the agent that mismatch "is the configured shape, not a bug". The precondition is routine (token expiry, `logout net`, authenticating `neko` before `net` during setup), and because `_try_single_cached_token` catches bare `Exception`, one transient error from the token endpoint at startup flips the identity for the whole process. Fix: fail closed for data capabilities (raise `AuthRequiredError(default_account)`); if identity tools should still answer, keep a separate identity-only fallback for `capability is None`; and flip `test_try_cached_token_loads_each_account`, which currently pins the leaky behaviour as expected.

**3. Scope.** Four things are bundled: routing, the aggregate tools, a 582-line README_CN, and a ROADMAP edit. Please split. I'll take **auth + routing + the gate** as one PR. The three `_all` tools cost every client ~464 proxy tokens per turn while refusing by default, and I have no demand signal for them yet — hold them. I can't review the translation and it isn't in the release checklist, so I'll decline it (it already copies one line that went stale).

Also needed in the auth/routing PR, detailed inline: routing should fail closed when the tool name can't be read from the SDK context, with one test through the real SDK; `get_account_credential` should surface `startup_error`; the new validators must not kill a working install at startup; an install that populated `accounts` on 1.21 shouldn't lose its login silently; `logout` semantics; the auth surfaces reflecting only the active account. Bookkeeping when you get there: README's category list still sums to 62, `SKILL.md` frontmatter, `ci.yml:60`'s message says "expected 62", and there's no CHANGELOG entry.

Suite runs green here (826 passed). Thanks again — the core of this is worth landing.

Nyaecho:
Thanks for the detailed review — you were right on all three counts. I've reworked this as a new branch containing **only the auth/routing/gate core**; the three `_all` aggregate tools, `aggregation.py`, README_CN, and the ROADMAP edit are all out (I'll bring aggregation back separately if/when there's demand, with the `except Exception`, `has_more`, and `count` clamping issues you flagged fixed first).

Point by point:

1. **One cache name, record-pinned identities.** All accounts now share `CACHE_NAME = "outlook-mcp"`; each account's `AuthenticationRecord` (its own `auth_record-<name>.json`, passed as `authentication_record=`) pins which identity the credential serves. The per-account cache-name helper is gone, README no longer claims per-name Keychain entries, and `cmd_logout` names the single shared entry. A comment at the constant explains the macOS constraint for the next person.
2. **Fail-closed startup.** `_active_account` is never re-pointed. When the default's token is missing, capabilities routed to it raise `AuthRequiredError('net')` with `outlook-mcp auth net` as the remedy; only identity tools (`capability is None`) fall back, via a startup-only `_identity_fallback_account` that is always logged. `switch_account` to an unauthenticated account clears the fallback — explicit actions get honest errors. The old test pinning the leak is inverted into `test_try_cached_token_is_fail_closed_when_default_fails`. The INSTRUCTIONS "not a bug" line is gone.
3. **Tightened catches.** The refresh path handles only `UnencryptedTokenCacheError`, the azure unencrypted-ValueError translation, `ClientAuthenticationError` (re-auth is the fix), and `AzureError` (transient — logged as such, no pointless re-auth advice); anything unexpected propagates. A transient error can no longer flip anything — nothing flips.
4. **`startup_error` in `get_account_credential`.** Checked before anything else, same as `get_credential`; a libsecret-less host reports the config fix on every routed tool.
5. **Logout.** Per-account record unlink; `credential` clears only when the active account is the one logged out (`logout("neko")` leaves `is_authenticated()` true for net); the CLI no longer lists per-account cache names or claims blanket re-auth — it says the logged-out account fails closed and others keep working. CHANGELOG records the record-deletion behavior change.
6. **Config validation is accept-and-warn** for everything 1.21 could have written (`default_account` without accounts, unusual names, duplicates, stale `default_account`, unknown capability keys); hard errors remain only where no legacy config can carry the mistake (`capability_accounts` naming an unknown account — a typo whose silent fallback would be identity substitution — and names that can't be filenames). `re.fullmatch` replaces the `$`-anchored match. The lifespan logs the offending fields and fails fast (booting on defaults would silently drop `read_only`); every CLI command, `logout` included, prints the fix and exits cleanly. An install that populated `accounts` on 1.21 also keeps its login: the legacy `auth_record.json` is adopted as the default account's record, with a warning.
7. **Routing fails closed.** `_calling_tool_name` accepts any `Mapping` (matching the SDK's type), and on a multi-account install an unreadable tool name — or a tool with no routing decision — raises `routing_unavailable` instead of routing to the active account. `test_real_sdk_client_gives_mail_and_todo_different_credentials` drives the server through a real `Client(mcp)` and asserts mail and todo receive different credentials under `{"mail": "net", "todo": "neko"}`, so the `ServerRequestContext.params` plumbing is exercised as shipped (verified against mcp 2.2.0).

Bookkeeping: the tool count stays 62 (no new tools; `switch_account` gained a parameter), so README's category list, SKILL.md, server.json, and ci.yml's "expected 62" guards are all correct as-is, and CHANGELOG has entries for the routing and the logout change. `whoami`/`auth_status` reflect the active account only, and `auth_status`'s remedy names it. Seven conventional commits on the branch; full suite and `ruff check` clean.

One mechanical note: the rework is pushed as `feat/multi-account-rework` on my fork (based on current main, 31a2d55), so this PR still shows the previous shape until it's re-pointed — happy to force-push it onto `feat/multi-account` here, or open a fresh PR and close this one, whichever you prefer.

mpalermiti:
Thanks for taking all three head-on — dropping the aggregates was the right call, and the record-pinned identities are exactly the fix.

**Open it as a fresh PR against current `main` and I'll take it from there.** I'm closing this one now with a pointer to it.

Reasoning, so it doesn't read as a brush-off: the rework isn't a revision of this PR, it's a different change — roughly half the diff is gone. My 11 inline comments here point at code that no longer exists, so force-pushing over the branch would leave a review thread that misleads whoever opens it in six months. I'm re-reviewing from scratch either way; a new PR makes that honest, and it gets a clean CI run instead of inheriting this one's state.

Two things for the rebase:

- I'm merging #62 today, so base on `main` after that lands — it adds `calendar_resolver.py` and touches README/SKILL, which is where the two of them overlapped.
- Point 1 is the one I care most about being pinned. The macOS cache-name collision is invisible in every test that doesn't have a real Keychain under it, so if there's a way to make `auth <b>` after `auth <a>` fail loudly in the offline suite rather than only in a live run, that's worth a commit of its own.

No rush on my end. And when you're ready to put up `feat/todo-checklist-and-attachments`, go ahead — I'd rather review two small ones than one large one.

PR #62, opened 2026-09-15
feat(calendar): let outlook_list_events read secondary calendars

Nyaecho:
## Summary

`outlook_list_events` always hit `/me/calendarView` — the **default** calendar — so events living in a secondary calendar (a class schedule, a shared team calendar) were unreachable: the tool returned an empty listing with no hint why.

This adds an optional `calendar` argument:

- omitted (or `"primary"`) → the default calendar, same request and same single round-trip as before — `/me/calendars` is never listed on this path;
- a **display name** → resolved case-insensitively against `/me/calendars`; not-found and ambiguous both raise with the calendars that *do* exist named in the error, so one `outlook_list_calendars` call away from a fix;
- a **Graph calendar ID** → passed through after syntactic validation, no listing call.

No new tools, no signature breaks, no query-string construction changes (`$filter`/`$orderby`/`$select` are built exactly as before — only the request builder's path changes), so no new live-tier guard is warranted under the CLAUDE.md rule. Tool count stays 62; CI/doc sync targets unaffected.

Complementary to #61 (multi-account): both touch `server.py`/`calendar_read.py` in different regions; this branch is based on `main` and rebases cleanly either way.

## Testing

- 6 new tests in `test_calendar_read.py::TestCalendarSelection`: omitted-keeps-default-path (asserts `/me/calendars` is *never* listed), `"primary"` alias, display-name resolution queries that calendar's view, Graph ID passes through without listing, not-found names what exists, ambiguity demands an ID.
- Verified against a real mailbox: a secondary calendar holding 20 events returned them all with `calendar: "<name>"`; the default path is byte-identical to before.
- Full offline suite green (the 7 Windows chmod/symlink failures are pre-existing on `main`).

## 中文说明

`outlook_list_events` 此前只查 `/me/calendarView`——即**默认日历**。日程建在自建日历里（课程表、共享团队日历）的用户会得到一个空列表，且没有任何提示。

本 PR 增加可选 `calendar` 参数：省略（或 `"primary"`）走默认日历，路径与请求数与从前完全一致（此路径不调 `/me/calendars`）；传**显示名**则大小写不敏感地在 `/me/calendars` 中解析，找不到或重名时报错并列出现存日历（离修复只差一次 `outlook_list_calendars`）；传 **Graph ID** 则校验后直通、不做列表调用。

不新增工具、不改签名、不改任何 `$filter`/`$orderby`/`$search` 构造（仅请求构建器的路径不同），按 CLAUDE.md 规则无需新增 live 层守卫；工具数保持 62，CI 与文档同步点不受影响。与 #61（多账号）互补，两者在 `server.py`/`calendar_read.py` 的改动区域不重叠，先后合并均可。

测试：`TestCalendarSelection` 6 个新用例（默认路径零额外往返、primary 别名、名称解析、ID 直通、未找到列出现存、重名要求 ID）；真实邮箱验证：自建日历 20 条事件全量返回，默认路径与从前逐字节一致；离线全量绿（7 个 Windows chmod/symlink 失败为 main 上的预存问题）。


mpalermiti:
Thanks — this closes a real gap (my own agent listed calendars ~60 times last month with no way to read the non-default ones), and the shape was right. I reviewed it in depth, ran the suite, and pushed **7b1bcd0** to your branch with the fixes below rather than round-tripping (maintainer edits were on — thank you). Please sanity-check it; if you'd rather rework any of it yourself, say so and I'll drop the commit.

**What changed in 7b1bcd0**

1. **Names before IDs.** `_looks_like_graph_id` was tuned for mail folders: any of `= + /`, or ≥40 chars, meant "ID". Ordinary calendar names trip that — `Kids + School`, `Work/Personal`, and the auto-generated `Calendar - Jane Smith (jane@…)` — so they never reached the name lookup and came back as "invalid characters" or a bare Graph 400. Every non-default reference is now matched against `/me/calendars`, first as an exact ID, then as a name (casefold + NFC). An ID that isn't one of the user's calendars is refused with the real list instead of going to Graph.
2. **The cursor carries the calendar.** It held only `$skip`, so page two without `calendar` silently paged the *default* calendar (200, plausible `has_more`, no error), and page two with it paid a second `/me/calendars` round-trip. The resolved id (None for the default) is stored in the cursor and validated on the way back in — it's caller-held state that goes into a URL path. Cursors minted before this still work with an explicit `calendar`. The cursor is decoded before any network call now, and base64's "Incorrect padding" is reported as "Invalid pagination cursor".
3. **`/me/calendars` is paged** (`$top=100`, `@odata.nextLink` followed) via a shared `calendar_resolver.fetch_all_calendars` that `outlook_list_calendars` now uses too. Graph doesn't document a page size for this collection, only that any collection may page.

Smaller things in the same commit: Graph-sourced names go through `sanitize_output` before they enter an error (parity with `outlook_list_calendars`); a matching calendar with no id is an error rather than a silent fall-through to the default (None was doing double duty); the ambiguous-name error lists the candidate ids; `""`, `"  "` and `"primary"` are one rule; single `CalendarViewRequestBuilder` import (the two SDK query-parameter classes are field-identical and kiota reads them off the object); docstring trimmed — every client pays for it each turn (surface 11,291 → 11,267 proxy tokens) — and a line in INSTRUCTIONS so the model passes calendar names directly instead of listing first; CHANGELOG entry crediting you; the `"Calendar" in message` assertion made non-vacuous and the secondary path's `request_configuration` inspected.

On the roadmap, not blockers: `create_event`, `list_events_delta`, `changes_since` and the `morning_brief` prompt are still pinned to the default calendar.

Suite: 821 passed locally. Security pass: clean — `validate_graph_id` plus kiota's RFC 6570 encoding keep an id inside `/me/calendars/{id}/`, and the cursor is a `$skip` int, not a URL. Merging once CI is green.

Nyaecho:
Sanity check done — 7b1bcd0 looks good from my side, thanks for pushing the fixes directly.

I verified each of the three changes against the code and with targeted probes: name-before-ID resolution (including "Kids + School"-style and auto-generated shared-calendar names, NFC/NFD and casefold variants), the calendar-pinned cursor (legacy cursors with an explicit `calendar` still work; tampered calendar values in the cursor are refused before any network call), and the paged `/me/calendars` fetch shared with `outlook_list_calendars`. No issues found, so no changes from me — the branch is clean as pushed.

Suite: 820 passed + 1 skipped on Linux-equivalent counts (the calendar/pagination/user tests all green locally; the only failures on my Windows box are the pre-existing symlink/permission-mode environment tests, unrelated to this PR), and `ruff check` is clean. Good to merge once CI is green.

Issue #41, opened 2026-09-07
outlook_create_event ignores recurrence — recurring series cannot be created


Wermeling:
## Bug Description

`outlook_create_event` ignores the `recurrence` argument entirely — recurring (series) events cannot be created through the MCP server. The parameter is accepted by the tool schema, but never applied to the Graph event, so every "recurring" event is created as a single occurrence with `recurrence: null`.

Verified on **outlook-graph-mcp 1.12.0 AND 1.14.0** (latest), mcp SDK 2.x, personal `@outlook.com` account.

## Steps to Reproduce

1. Call `outlook_create_event` with a recurrence payload:
   ```json
   {
     "subject": "Weekly sync",
     "start": "2026-09-07T12:30:00+02:00",
     "end": "2026-09-07T13:30:00+02:00",
     "recurrence": {
       "pattern": {"type": "weekly", "interval": 1, "daysOfWeek": ["monday", "friday"]},
       "range": {"type": "noEnd", "startDate": "2026-09-07"}
     }
   }
   ```
2. Server responds: `Pydantic validation error` — `recurrence: Input should be a valid string` (the bridge passes the parsed dict, but the schema declares `recurrence: str | None`).
3. Workaround attempt: pass `recurrence` as a JSON **string** instead.
4. Server responds `status: created`, but the event is a **single occurrence** — fetching it via `outlook_get_event` (or Graph `GET /me/events/{id}`) shows `recurrence: null`, `type: occurrence`.

## Root Cause

In `outlook_mcp/tools/calendar_write.py` (`create_event`, ~line 81), the `recurrence` parameter is received but **never assigned to the Graph event** — `event.recurrence` is never set before `graph_client.me.events.post(event)`. Additionally the tool schema in `server.py` types `recurrence: str | None`, so dict payloads are rejected by Pydantic before reaching the handler.

## Expected Behavior

Passing a recurrence (dict or JSON string) should create a real Outlook series (`type: seriesMaster` with `recurrence` populated), e.g.:

```json
{
  "recurrence": {
    "pattern": {"type": "weekly", "interval": 1, "daysOfWeek": ["monday", "tuesday"]},
    "range": {"type": "noEnd", "startDate": "2026-09-07"}
  }
}
```

## Actual Behavior

- Dict → Pydantic validation error (`Input should be a valid string`)
- JSON string → single occurrence created, `recurrence: null` in the stored event
- No path currently creates a recurring series via the MCP

## Environment

[...]

PR #63, opened 2026-09-17
fix(contacts): return the fields Graph already sends, and write them back

neilbrencode:
`outlook_get_contact` sends no `$select`, so Graph returns the whole contact — and the detail formatter read 12 fields of it. A contact with a home address, two categories and a note read back as having none of them, which from the caller's side is indistinguishable from a contact that genuinely has none. The data was never missing; the formatter dropped it.

Three commits: the read fix, the write half that makes an address correctable rather than merely visible, and the live guards for both.

## What changed

**Read.** `outlook_get_contact` now returns `home_address`, `business_address`, `other_address`, `categories` and `personal_notes`. Graph sends an empty `physicalAddress` object rather than null for an address a contact does not have, so an empty one is reported as `None` — otherwise every contact appears to have three blank addresses.

**`$select` audit on the two listings**, which found both ends of the same mistake: `givenName`, `surname` and `title` were selected and never read; `categories` was read by nobody because it was never selected. Both lists were duplicated verbatim and are one constant now. Categories are on the listing only — Graph's `$search` over contacts does not return them, so the search path omits the key rather than reporting every contact as uncategorised. Absent and empty mean different things.

**Write.** `outlook_update_contact` gains `home_street`, `home_city`, `home_state`, `home_postal_code`, `home_country`. Supplying none of them leaves the stored address untouched.

## The bug the live tier found, which no offline test could

The partial case — someone knows the city and not the street — **400'd**. Assigning `None` to the parts not supplied looked equivalent to leaving them unset and is not: the Graph request adapter serializes through the backing store, which emits an explicitly-`None` field, and for a nested model emits it onto the *parent* under its Python name. Captured off the wire:

```
PATCH /me/contacts/{id}
{"@odata.type": "#microsoft.graph.contact", "country_or_region": null, "postal_code": null,
 "state": null, "street": null, "homeAddress": {"city": "Bothell"}}      -> 400
```

> The property 'country_or_region' does not exist on type 'microsoft.graph.contact'.

The full five-part write returned 200 throughout, so the feature worked for the case anyone would try by hand and failed for the case it was written for.

Nothing offline saw it. The mocked suite asserts the model; `test_write_payloads_reach_the_wire.py` serializes with a bare `JsonSerializationWriter`, which omits `None`-valued fields exactly as its docstring says. So the offline regression here goes through `BackingStoreSerializationWriterProxyFactory` — the adapter's own serializer — and asserts a city-only patch carries no nulls at all. I checked it the way it deserves: it **fails against the previous version of the commit** and passes with the fix.

Two consequences in the code: parts are assigned one at a time rather than through a `setattr` loop, because `test_sdk_fields_exist` resolves `<var>.<attr> = ...` statically and cannot see `setattr` — a loop would have moved the five field names out from under the guard built for that class of bug. And `_build_address` sets only what it was given.

## Graph replaces the address; it does not merge

Verified live: patch a city onto a stored address and the street, state and postal code come back empty. Same shape as `attendees` on `outlook_update_event`, so it is documented the same way — in the tool docstring the model reads, and in the README row — rather than left as a surprise. The read fix is what makes that workable: the caller can read the address back and pass the parts it means to keep. A live guard pins the replacement and fails if Graph ever starts merging.

## Verification

- `uv run pytest` — 808 passed, 1 skipped. The 4 failures are the pre-existing POSIX mode-bit tests (`test_attachment_paths.py`, `test_config.py`), unchanged on `main`.
- `uv run pytest -m live` — 16 passed, including the two new contact `$select` guards. Both hunt their fixture across pages: contacts come back ordered by `displayName`, and on my mailbox the first 100 have no categories at all, so a page-one assertion would have skipped forever while reading as covered.
- `OUTLOOK_MCP_LIVE_WRITE=1 uv run pytest -m live_write` — 19 passed, 1 failed, the failure pre-existing (below). New: full address, partial address, replacement semantics, and a name-only update that must leave the address alone.
- One real round trip through the MCP server itself, not the tool functions — create → full address → partial → `get_contact` → `list_contacts` → delete — since `server.py` passes these positionally and the tool-level tests cannot see a mis-ordered argument.
- `ruff check` clean. `ruff format` leaves `contacts.py` and `test_live_contacts_write.py` clean; `test_live_query_shape.py` and `test_write_payloads_reach_the_wire.py` were already unformatted on `main` and I left them alone.
- Tool surface **11,386 → 11,566** proxy tokens (+180, +1.6%; ceiling 12,000). `outlook_update_contact` 193 → 374, nearly all of it the five parameters; the docstring is two lines because a warning about silent data loss has to be where the model reads it.

## Scope, and what I deliberately did not do

- **`outlook_create_contact` takes no address.** It would cost roughly the same again in schema tokens, and an address is usually filled in later — but say the word and it is a small commit.
- **`businessAddress` and `otherAddress` read back but stay unwritable.** Home is the one a contact list needs; guessing at the other two without a caller is how a surface bloats.
- **No way to clear an address.** Whitespace-only parts count as absent, so the clearing idiom would have to be explicit (`reply_to=[]` is the precedent). Held until someone wants it.
- **`home_country` vs the response's `country_or_region`.** I kept the friendlier parameter name and matched Graph on the way out. Happy to rename either side for exact symmetry if you would rather they read identically.
- **CHANGELOG** entry is under `## [Unreleased]`, alongside #60's — move or rename it at release as you see fit.

## Two things that are not mine to fix

- `test_live_calendar_write.py::test_online_meeting_is_not_supported_on_personal_accounts` **fails on `main`** as of today: `is_online` comes back `True` after the patch, so whatever Graph was doing with `isOnlineMeeting` on personal accounts has changed. Unrelated to this PR; flagging so the one red test in the write tier does not look like mine.
- `test_write_payloads_reach_the_wire.py`'s premise — "the SDK omits None-valued fields" — is true of `JsonSerializationWriter` and not of the path the adapter takes. `remove_recurrence`'s `additional_data` workaround is unaffected and still correct, but the file's `wire()` helper cannot see a field assigned `None`. Switching it wholesale is a separate change with its own blast radius, so I left it and scoped the proxy-based serializer to the contacts test that needs it.

<sub>*- Claude Code*</sub>

🤖 Generated with [Claude Code](https://claude.com/claude-code)


neilbrencode:
Please take a look when you get a chance @mpalermiti

mpalermiti (changes requested):
This is a good PR and the central find is better than the PR title suggests. The read fix is obviously right — a formatter narrower than the `$select` is a silent lie, and "absent and empty mean different things" is the correct instinct. But the partial-address 400 is the real contribution, and it generalizes well past contacts: I've pulled it into the shared guard as #64, details at the bottom.

Requesting changes on five things, then a handful in the new test helper that I'd rather fix now than have copied.

## Blocking

**1. The read and write halves disagree on a field name.** `_ADDRESS_FIELDS` ends in `country_or_region`; the parameter is `home_country`. You flagged this as a naming preference — it isn't, because of what the docstring tells the caller to do. "Pass every part you want to keep (`outlook_get_contact` returns them)" has exactly one mechanical follow-through:

```python
update_contact(**{f"home_{k}": v for k, v in detail["home_address"].items()})
```

Four of five resolve. `country_or_region` → `home_country_or_region` doesn't. A strict client errors on an operation the docstring just prescribed; a lenient one drops it, the write goes through, and since Graph replaces, the country is gone — the precise data loss the warning exists to prevent. `test_read_and_write_agree_on_the_same_five_fields` passes parts positionally, so it structurally cannot see this. Rename either side; I don't care which, but they have to match, and the test should compare names rather than positions.

**2. `_format_address` and `_build_address` disagree about whitespace.** `any(parts.values())` counts `" "` as content; `if value and value.strip()` at line 87 drops it. So a contact whose stored street is a stray space reads back as *having* an address, and handing those same five values straight back — the documented way to preserve one — produces `None`, skips the write, and still answers `{"status": "updated"}`. The two functions are documented as mirrors. `any(v.strip() for v in parts.values())` in both directions.

**3. An all-empty address write reports success having done nothing.** Same root as 2. `home_street=""` through `home_country=""` collapses to `present == {}` → `_build_address` returns `None` → the assignment is skipped → `{"status": "updated"}`. A model asked to clear an address will do exactly this and report success. Worse, four lines up `if last_name is not None:` means `""` *clears* the surname — so one signature has two opposite meanings for "provided," documented nowhere. I'm fine deferring the clearing capability. I'm not fine reporting a no-op as an update: either reject an all-empty address explicitly, or add the `remove_home_address: bool` flag mirroring `remove_recurrence`.

**4. `personal_notes` and the address parts need `multiline=True`.** `_CONTROL_CHARS` excludes `\x0d` and `sanitize_output(multiline=False)` only replaces `\n` and `\t`, so an Outlook note comes back as one run-on line with a stray CR — which a terminal client will use to overwrite the line. Every other body-shaped field in the repo passes `multiline=True` (`mail_read.py:248/339/400`, `calendar_read.py:176`, `todo.py:98`); these two are the only exceptions. For addresses it's not cosmetic: Outlook's Street box is multi-line, and the docstring tells the caller to read the parts back and write them in, so the flattened value is what lands in the mailbox. The new test at `tests/test_contacts.py:439` uses a single-line note, so nothing catches it.

**5. SKILL.md wasn't updated.** Line 119 is still `- outlook_update_contact — Update`. README got the three new facts, but README isn't in the wheel (`pyproject.toml` packages only `src/outlook_mcp`) and #60 just moved the install to PyPI — so for a skill consumer SKILL.md *is* the documentation. As it stands they never learn `home_*` exists and, more to the point, never learn about replace-wholesale, so their first partial `home_city` write silently blanks the street, state, postal code and country. Lines 115 and 117 want the categories asymmetry too.

## In the new helper — worth fixing before it gets copied

**6. `_adapter_wire` is destructive, and that makes it silently vacuous on a second call.** `on_after` sets `is_initialization_completed = True`, whose setter rewrites every entry as `(not value, val[1])` — clean — and recurses into nested models (`in_memory_backing_store.py:34-40`). So:

```python
assert '"street": null' in _adapter_wire(contact)
assert '"street": null' in _adapter_wire(contact)   # fails
```

Add that second line to `test_the_serializer_this_uses_is_the_one_that_can_see_it` and watch it go red. The next person who calls it once for a debug message and once to assert, or reuses a model across a parametrized case, gets a green guard for exactly the bug this PR exists to catch. Needs a fresh model per call, and the docstring needs to say so out loud.

**7. Point it at a top-level `None` and you get an opaque kiota error, not an assertion.** Same `on_start` path puts a top-level null on the root writer's `.writer` while the body goes to `.value`, and `get_serialized_content()` raises `ValueError("Invalid Json output")`. The first person to aim this at a "clear this field" body — `remove_recurrence`, or the clearing semantics from 3 — gets that where they expected an assertion, and the natural reaction is to fall back to the bare writer, which deletes the only serializer that can see the leak. Worth catching and re-raising as an assertion that says what to do instead.

**8. `assert "null" not in body` will redden for reasons unrelated to contacts.** Parametrize over realistic addresses and a street on *Nullah Road*, or a city named *Nulle*, fails a correct payload with the message "a part we were never given reached the wire." The two narrow assertions on either side already pin the real regression; `assert json.loads(body).keys() == {"homeAddress"}` is stricter and order-independent. Line 625 has the same shape and additionally depends on `json.dumps`' default separators and key insertion order.

**9. The live fixture's per-page assertion can't fail.** `assert all("categories" in c for c in page["contacts"])` asserts a key `_format_contact_summary(..., with_categories=True)` writes unconditionally. If Graph stopped honouring the `$select` tomorrow every contact would come back `"categories": []` and this still passes — while the docstring tells the next reader the fixture is what proves the key is real. Only `assert categorised_contact["categories"]` carries information. Two notes: it belongs in the test, not the fixture (an assertion in a fixture surfaces a product regression as a pytest ERROR on two tests, which reads as infrastructure breakage), and counting non-empty lists across the pages you already walk and *failing* beats skipping.

**10. The live search test can error on a punctuation-heavy contact.** It feeds `display_name.split()[0]` into `sanitize_kql`, which strips `" \ & | ! *` and raises `ValueError("Search query is empty after sanitization")`. A contact displayed as `*Mom*` or `!Emergency` errors the categories test for a reason that has nothing to do with categories.

## Not blocking

- **`business_address` / `other_address` readable but not writable** is a defensible line and I'd take the PR without them — but `_build_address`'s own docstring names readable-but-not-settable as the defect to avoid, so right now it reads as if the class of bug is closed when it isn't. One sentence saying home is deliberately first is enough. Same for `create_contact` taking no address: the two-call sequence isn't written down anywhere, and because the update *replaces*, getting one part wrong on the second call discards the rest.
- **The categories asymmetry is explained only in README, which no model reads.** "Omit rather than lie" is worth exactly as much as the caller's ability to know it — and the caller is an LLM reading `inputSchema`. An agent that searches, sees no `categories`, and has nothing in context will report "no categories," which is the confusion you set out to kill, moved up a level. One clause in the two docstrings or one line in INSTRUCTIONS.
- **`if email:` vs `if email is not None:`** (lines 332 / 345) — pre-existing, but it's inside the function you rewrote and it's a third absent-vs-empty rule in one signature. `email=""` skips validation and still reaches Graph.
- **`categorised_contact` runs its 10-page walk twice** — function-scoped with two consumers, up to 20 round trips per `-m live`. The obvious fix doesn't work (`real_graph_client` is deliberately function-scoped, `conftest.py:88-100`, so `scope="module"` raises ScopeMismatch); cache the dict in a module-scoped fixture or fold the two assertions together.
- **`_ADDRESS_FIELDS` is used only by `_format_address`** while `_build_address` re-spells the same five names three more times; the categories comprehension is duplicated at `contacts.py:146` and `:178`; `_format_address`'s `if address is None: return None` is dead (`getattr(None, "street", "")` is `""`, so the emptiness check below already returns `None`); `test_read_and_write_agree_on_the_same_five_fields` is `async def` and awaits nothing.

## What I've taken off your plate

**#64 generalizes your find into the shared guard.** You were right that `test_write_payloads_reach_the_wire.py`'s premise is false for the adapter's path, and right to scope around it rather than rewrite it mid-PR. I've done the rewrite: `wire()` now goes through the backing-store proxy, and `assert_on_wire` additionally rejects any snake_case key (Graph is camelCase, so an underscore key is always a leak) and any null we didn't ask for. Verified by mutation — injecting `event.online_meeting_url = None` into `update_event` turns the file red, and the same mutation passed 14/14 before.

The quieter half of your discovery is what convinced me it belonged there: contacts only 400'd because `country_or_region` is multi-word. There are ~40 single-word assignments in `tools/` where the Python name *is* a real Graph property — a leak there gets accepted and clears the field.

**If #64 lands first, import `wire` from the family file and drop `_adapter_wire`** — items 6, 7 and 8 are already handled there.

Two follow-ups that are mine, not yours: `_format_contact_delta` still drops `categories` (same bug one module over, and its "mirrors field-for-field" docstring is now false while SKILL.md tells agents to prefer the delta tool), and the mail summary `$select` is copy-pasted across three modules with only `mail_read.py:141` including `inferenceClassification` — so three of four paths return a permanently-empty `classification`. Your own comment describes that defect exactly.

Nice work. The five blockers are all small; the helper items are the ones I'd not skip.

mpalermiti:
One addition to the review — small, and it only becomes true when this merges, which is why I'm putting it here rather than fixing it myself.

`_format_contact_delta` (`contacts_delta.py:50`) says it "mirrors `contacts._format_contact_summary` field-for-field so callers don't have to branch." That's accurate on `main` today — both return the same five keys. This PR adds `categories` to the summary and not to the delta, so merging it makes the docstring false.

That matters more than a stale comment, because `server.py:916` and `SKILL.md:121` both steer agents to the delta tool for recurring work. An agent that seeds from `outlook_list_contacts` and refreshes from `outlook_list_contacts_delta` either `KeyError`s on `c["categories"]` or quietly reports every changed contact as uncategorised — which is the same "empty means absent" lie this PR exists to kill, one module over.

It should be cheap: `/me/contacts/delta` accepts no `$select` at all (your comment at `contacts_delta.py:96` notes Graph 400s if you pass one), so Graph is already sending `categories` in the raw payload and line 62 is dropping it. One key, and the parity claim holds again.

If you'd rather keep this PR's diff where it is, say so and I'll take it in a follow-up — but I'd rather it land with the change that causes it than trail behind it.

Two related things landed while you were working, both from your find:

- **#64** rewrote `test_write_payloads_reach_the_wire.py` to serialize through the backing-store proxy, and `assert_on_wire` now also rejects snake_case keys and unrequested nulls. Items 6, 7 and 8 in my review are handled there — rebase and you can drop `_adapter_wire` and import `wire` instead.
- **#65** is the same defect class as yours in mail: four copies of one `$select`, only one of which asked for `inferenceClassification`, so three tools reported every message as unclassified. The guard there reads the formatter's AST and derives what the `$select` must contain. **There's a `_PAIRS` table in `tests/test_select_covers_the_formatter.py` that contacts should join once this lands** — it would have caught your original bug, and it'll catch the next one. Worth a row for `_format_contact_summary` and `_format_contact_detail` in this PR if you're willing; otherwise I'll add them after.

neilbrencode:
All fifteen addressed, in two commits on top so you can read just those. Rebased onto `main` with #64 and #65 in.

## Blocking

**1. The name mismatch.** You were right that it isn't a naming preference, and the fix follows from why: if the docstring's instruction is "hand the address back", then the address is the argument. `home_address`, `business_address` and `other_address` each take the shape `outlook_get_contact` returns, so the round trip is `home_address={**stored, "city": "Issaquah"}` rather than a five-key rename. `_ADDRESS_FIELDS` is now the single tuple both halves are built from, and `TestTheReadAndWriteHalvesNameTheSameParts` compares key *sets* — including one test that runs the documented sequence literally, and one that asserts the read path emits exactly the keys the write path accepts. The live tier performs the same round trip against Graph.

**2. Whitespace.** `any(value.strip() for value in parts.values())` in the formatter, matching the builder. The case you described — stored street of a stray space, read back as real, handed straight back, refused — is pinned in both directions.

**3. An all-empty write.** It's an error now, not a no-op reporting success: *"home_address has no content. Omit it to leave the stored address unchanged — this tool cannot clear an address."* I took reject over `remove_home_address` because the flag costs surface on every turn to advertise a capability nobody has asked for, while the error costs nothing and says the same thing at the moment it matters. Parametrized over `{}`, empty strings, whitespace and explicit nulls. An unknown key is refused the same way, naming the valid parts — a misspelt part that patched nothing is #41's shape. And `if email:` → `if email is not None:` while I was in there, so `email=""` no longer skips validation and reaches Graph as a blank address.

**4. `multiline=True`.** On `personal_notes` and the address parts. Both guards fail against the unfixed formatter — I mutated it and watched them redden, because a sanitization test that passes either way isn't one. Your reason for addresses is the sharper one and it's in the test name: the docstring tells the caller to write these values back, so a flattened street is what lands in the mailbox.

**5. SKILL.md.** All four contact lines: what `get_contact` returns, the categories asymmetry on both listing tools, the address shape with replace-wholesale, and a note on `create_contact` that an address is a second call. Point taken that README isn't in the wheel — that was the gap worth more than the text.

## The helper

**6, 7, 8** are gone with `_adapter_wire`: #64 landed the same helper in the family file with the single-use behaviour and the top-level-null case documented, so `test_contacts.py` imports `wire` and keeps only what's address-specific — the partial write asserted key by key (`set(body) == {"@odata.type", "homeAddress"}`), which is where the leak actually happened, plus the same for a business-only write. Your `assert_on_wire` upgrade covers the rest of it for every tool, which is the better place for it.

**9.** The vacuous per-page assertion is deleted rather than moved — you're right that it asserts a key the formatter writes unconditionally, so it could never fail. The walk returns a tally, the tests judge, and it runs once per session now (module-level cache; `real_graph_client` being function-scoped rules out `scope="module"`, as you said).

One thing I did differently, and say so rather than quietly: I kept the **skip** when a mailbox has no categorised contact, with the tally in the message (`walked <n> contacts, none categorised — cannot tell an honoured $select from an ignored one`). Failing there would break the module's own contract — "skip cleanly when the mailbox lacks the data a given assertion needs. They must never assume this maintainer's mailbox" — for a mailbox that is simply not using categories, which is not a regression in anything. What I took from your point is that a silent skip is the real problem, so the skip now reports what it walked. If you'd rather it fail, that's a one-line change and I'll make it.

**10.** `_searchable()` keeps only alphanumeric characters and needs three of them, so `*Mom*` and `!Emergency` can't reach `sanitize_kql`; it walks the candidates and skips only if none has a usable fragment.

## Not blocking

- **`business_address` and `other_address` are writable.** You'd have taken the PR without them, but the measurement changed the calculus: fifteen flat parameters cost **644** proxy tokens, three objects cost **336**. Three addresses for less than the one address cost before — and the object shape is what fixes blocker 1, so the two decisions turned out to be the same decision. `outlook_update_contact` goes 193 → 372 and the surface 11,449 → 11,628, against the 12,000 ceiling.
- **`create_contact` still takes no address.** The two-call sequence is written down now — INSTRUCTIONS and SKILL.md — including that the update replaces, so getting one part wrong on the second call discards the rest.
- **The categories asymmetry** went into INSTRUCTIONS rather than the two docstrings: it is guidance that spans three tools, so it's paid for once per session instead of every turn. It says not to report a searched contact as uncategorised, and where to read them from instead.
- Also taken: `_ADDRESS_FIELDS` now feeds both halves rather than being re-spelled; `_categories` deduplicates the comprehension; the dead `if address is None` is gone; the round-trip tests aren't `async` any more.

## Your follow-ups

**The delta.** Taken in this PR — you're right that it should land with the change that causes it. `categories` was already in the payload (`/me/contacts/delta` takes no `$select`) and line 62 was dropping it. The parity claim is a test now rather than a sentence: `test_both_summaries_carry_the_same_keys` compares key sets, so the next field added to either side fails until it's added to both. Mutation-checked — dropping the line reddens four tests.

**`_PAIRS`.** `_format_contact_summary` has a row, paired with `_LIST_SELECT` (the widest select that formatter can be fed; the search path's narrower one is honoured by omitting the key). Two things had to change first, both worth your eye:

- The guard now follows a call to a same-module helper handed the model as its first argument. Without it the summary's phone (`_primary_phone`) and categories (`_categories`) reads are invisible, and the row *passes both directions* while reporting three selected fields as unread — the guard going quietly blind in exactly the way it exists to prevent. `test_a_field_read_through_a_helper_still_counts_as_read` pins it.
- `_primary_phone` now spells out `home_phones` and `business_phones` instead of looping over a tuple of names. A `getattr` whose attribute is a variable can't be read off the AST, so the loop was holding two fields outside the guard.

`_format_contact_detail` doesn't have a row, and I don't think it can have an honest one as things stand: its `$select` is the empty one, and a row pairing a formatter against a select nobody sends asserts nothing. What I pinned instead is the contract that makes the formatter safe — `TestGetContactAsksGraphForEverything` asserts the request carries no `request_configuration` and that the detail shape is all seventeen keys. If you'd rather `get_contact` sent an explicit select derived from the formatter, the row becomes real and the response gets smaller; the guard makes that safe to do. Say which and I'll do it here.

## Verification

- `uv run pytest` — 874 passed, 1 skipped. The 4 failures are the POSIX mode-bit tests, unchanged on `main`.
- `uv run pytest -m live` — 16 passed, both contact `$select` guards among them, non-vacuously — page one of this mailbox has no categorised contact at all, which is what the paging walk is for.
- `OUTLOOK_MCP_LIVE_WRITE=1 uv run pytest -m live_write` — 20 passed, 1 failed: `test_online_meeting_is_not_supported_on_personal_accounts`, which **fails on current `main` too** — `is_online` now comes back `True` after the patch, so Graph's behaviour for `isOnlineMeeting` on consumer mailboxes appears to have changed since that guard was written. Not mine, flagging so the red doesn't look like it.
- One real round trip through the MCP server itself: create → three addresses → read → hand the address back with one part changed → delta → delete. And the refusals through the real client, because "reaches the model" is the whole point of the error hierarchy: `Error executing tool outlook_update_contact: home_address has no content...` — text intact, not the withheld form. That check also found that a non-dict argument is rejected by the generated schema before `_build_address` sees it, so that branch is defensive for direct library callers only.
- Input errors are plain `ValueError` again, which is what every other tool module raises and what `_wrap_tool_errors` expects; `contacts.py` was the only module importing `ToolInputError` directly.
- `ruff check` clean.

<sub>*- Claude Code*</sub>

mpalermiti (approved):
All fifteen are addressed. I checked it independently rather than on the summary — 879 passed offline, `ruff check` clean, tool surface inside the ceiling. Two commits on top was the right call; it made this readable.

**The object shape is better than what I asked for, and I want to say why rather than just take it.** I asked you to rename a field. You noticed that the rename was a symptom — if the docstring's instruction is "hand the address back," then the address is the argument — and the mismatch stops being possible rather than being corrected. That all three addresses became writable for less than the one flat address cost is the part I'd have bet against; `business_address` and `other_address` were on my "defensible to defer" list precisely because I assumed they'd cost. Measuring beat my guess. Taken, including the `_ADDRESS_FIELDS` single tuple and the key-set comparison, which is the test that could actually have caught the original.

## One correction, and it's in a docstring

Your reasoning for following helper calls is right. Your account of what happens without it isn't, and since it's written into `_fields_read_from`'s docstring it'll be the next reader's mental model.

The docstring says that without the extension the guard "would report three selected fields as unread and miss the one that matters," and the PR comment says the row *passes both directions*. It doesn't pass. It fails:

```
contact summary: the $select asks for ['businessPhones', 'categories', 'homePhones',
'mobilePhone'], which _format_contact_summary never reads. Drop them, or read them.
```

I then ran the case that matters — the extension removed *and* `categories` dropped from `_LIST_SELECT`, which is the original bug — and got:

```
without the extension: the $select asks for ['businessPhones', 'homePhones', 'mobilePhone'],
                       which _format_contact_summary never reads. Drop them, or read them.

with the extension:    _format_contact_summary reads ['categories'], which the $select never
                       asks Graph for.
```

So the failure mode isn't silence, it's **a loud failure pointing the wrong way**. Someone acting on the first message deletes three phone fields from the `$select`, breaks phone on the listing, and leaves `categories` exactly as broken as it was — the guard actively steering the fix into a second bug. That's worse than blind, and it's a better argument for your change than the one you made.

Worth getting right on its own terms, but especially here: #64 existed because `calendar_write` recorded the wrong *reason* for a correct workaround, and the wrong reason nearly got the workaround deleted. Same shape. One line.

Everything else in the helper extension I'd keep as-is — same-module only, first-argument only, `_seen` guarding recursion, and spelling out `_primary_phone`'s fields because a `getattr` with a variable attribute can't be read off the AST. That last one is a real constraint on this style of guard and it deserved the comment you gave it.

## Your two questions

**The skip stays.** You're right and I was imprecise. My objection was to a *silent* skip, not to skipping — a live tier that assumes my mailbox is a live tier that only works on my mailbox, and that's a rule worth more than this assertion. The tally in the message is the fix; it distinguishes "nothing to check" from "checked nothing," which is all I actually wanted.

**Don't give `get_contact` a `$select`.** You talked yourself out of it and you were right, so let me make it explicit: it sends none today, so Graph returns the whole resource and the formatter can never be starved — the bug class is *impossible* on that path. Adding a select to make a `_PAIRS` row possible would create the failure mode the row would then be guarding against. That's optimizing for coverage rather than for correctness. `TestGetContactAsksGraphForEverything` pins the actual contract and is the right guard. If the response size ever justifies a select, it's a separate change with its own live check.

## `is_online`

Second independent sighting, so it's probably real — I'm not treating it as confirmed until it's checked deliberately, and that's mine to do, not yours. Thanks for flagging it twice rather than assuming I'd seen it.

---

Fix the docstring and I'll merge. Nice work — the shape change in particular is the sort of thing I'd rather be talked into than not.

neilbrencode:
Fixed in 995d961 — and you're right on the substance, not just the wording. I reproduced both directions rather than taking the correction on trust, and the captured output is now in the docstring verbatim:

```
A. extension removed, $select intact
   the $select asks for ['businessPhones', 'categories', 'homePhones', 'mobilePhone'],
   which _format_contact_summary never reads. Drop them, or read them.

B. extension removed, categories dropped from _LIST_SELECT (the original bug)
   the $select asks for ['businessPhones', 'homePhones', 'mobilePhone'],
   which _format_contact_summary never reads. Drop them, or read them.

C. extension present, categories dropped from _LIST_SELECT
   _format_contact_summary reads ['categories'], which the $select never asks Graph for.
```

B is the one that matters: with the real bug present and the extension absent, the guard reports the three phone fields and says nothing about `categories`. Someone following it deletes phone from the `$select` and leaves the bug — which, as you say, is worse than silence, because silence doesn't hand you a wrong fix with a passing test at the end of it. The docstring now carries the consequence, not just the mechanism.

My PR comment above has the same error in it — "the row *passes both directions*" is wrong, and this is the correction for that too. The previous commit message says it wrong as well; rather than rewrite a commit you've already reviewed, 995d961 states the correction in its body.

Everything else left as you described: same-module only, first-argument only, `_seen` guarding recursion, `_primary_phone` spelled out, `get_contact` keeps its empty `$select`, the skip keeps its tally.

`ruff check` clean, 874 passed here (the 4 reds are the POSIX mode-bit tests — Windows). The guard's code is untouched; this is a docstring.

<sub>*- Claude Code*</sub>

PR #66, opened 2026-09-17
feat: multi-account auth with per-capability routing and a cross-account gate

Nyaecho:
Supersedes #61 — same core, reworked to the reviewed shape. This PR contains only the auth/routing/gate core: the three `_all` aggregate tools, `aggregation.py`, the Chinese README translation, and the ROADMAP edit are all out. Based on current `main` (post-#62; the only textual overlap was in the docs, and both sides are kept).

## What this adds

- `accounts[]` with per-account device-code login (`outlook-mcp auth <name>`), per-account `AuthenticationRecord`s over **one shared token cache** — `CACHE_NAME` is the same for every account, and `authentication_record=` pins which identity each credential serves. A comment at the constant documents the macOS constraint (the Keychain item is fixed; the name only picks the signal file).
- `capability_accounts: {"mail": "net", "todo": "neko", …}` routes each capability group to an account; `allow_cross_account` (default `false`) gates `switch_account(name, capability?)`.
- **Fail-closed startup.** When the default account's token is unavailable, capabilities routed to it raise `AuthRequiredError(<account>)` with the `outlook-mcp auth <account>` remedy; only identity tools (`capability is None`) fall back, via a startup-only `_identity_fallback_account` that is always logged and cleared by an explicit `switch_account`. `_active_account` is never re-pointed.
- **Tightened exception scope** in the cached-token path: `UnencryptedTokenCacheError`, the azure unencrypted translation, `ClientAuthenticationError` (re-auth is the fix), and `AzureError` (transient — logged as such, no pointless re-auth advice). Anything unexpected propagates; a transient token-endpoint error can no longer flip anything.
- `get_account_credential` surfaces `startup_error` before anything else, same as `get_credential` — a libsecret-less host reports the config fix on every routed tool.
- **Logout.** Per-account record unlink; `credential` clears only when the active account is the one logged out (`logout("neko")` leaves net authenticated). The CLI says the logged-out account fails closed and the others keep working. CHANGELOG records the record-deletion behavior change.
- **Config validation is accept-and-warn** for everything 1.21 could have written (`default_account` without accounts, unusual names, duplicates, a stale `default_account`, unknown capability keys); hard errors remain only where no legacy config can carry the mistake (`capability_accounts` naming an unknown account, names that can't be filenames). `re.fullmatch` replaces the `$`-anchored match. Lifespan and every CLI command — `logout` included — print field-level fixes instead of dying bare, and a pre-multi-account `auth_record.json` is adopted as the default account's record with a warning, so an upgrade doesn't lose the login.
- **Routing fails closed** when the tool name can't be read from the SDK context (`Mapping` accepted, matching the SDK's type): on a multi-account install an unreadable name raises instead of routing to the active account. `test_real_sdk_client_gives_mail_and_todo_different_credentials` drives the server through a real `Client(mcp)` and asserts mail and todo receive different credentials under `{"mail": "net", "todo": "neko"}`.

## The offline Keychain-collision test you asked for

`tests/test_keychain_collision.py` simulates the macOS persistence semantics offline: a single shared storage slot that every save replaces wholesale, with the name only gating a signal file. `auth net` then `auth neko`, then a fresh `AuthManager` (a server restart) — both accounts silently authenticate, every write uses the one `CACHE_NAME`, and net's token is still net's. Reintroducing a per-account cache name fails both tests loudly (verified by flipping the name locally: the written-names assertion and the record-pinned lookup both go red, the latter naming the per-account cache it can't find). Separate commit, as requested.

Tool count stays 62 — no new tools; `switch_account` gained a parameter — so README/SKILL/server.json and the ci.yml guards are unchanged and correct. CHANGELOG has entries for the routing and the logout change. Suite: 873 passed locally (7 pre-existing Windows-only environment failures, unrelated); `ruff check` clean. Eight commits.

## 中文说明

替代 #61：同一核心按评审意见重做，只保留 auth/路由/门控——三个 `_all` 聚合工具、`aggregation.py`、中文 README、ROADMAP 改动全部移出；基于合并了 #62 的当前 `main`。要点：**单一共享令牌缓存**（全账号共用一个 `CACHE_NAME`，每账号 `AuthenticationRecord` 钉定身份，规避 macOS Keychain 条目固定导致的互相覆盖）；**按能力路由**（`capability_accounts` + `allow_cross_account` 总闸门控 `switch_account`）；**启动 fail-closed**（默认账号令牌失效时路由到它的能力抛 `AuthRequiredError` 并给出 `auth <账号>` 补救，`_active_account` 永不被改写，仅身份工具有带日志的启动期回退）；缓存加载异常按语义分类捕获、瞬时错误不再翻转身份；`get_account_credential` 先报 `startup_error`；**logout 按账号解绑**（登出非活跃账号不影响其余账号）；**配置校验对 1.21 旧形态 accept-and-warn**（升级不启动即死、旧登录自动采纳），硬错误仅限旧配置不可能出现的笔误；**路由读不出工具名时 fail-closed**（收 `Mapping`，真实 `Client(mcp)` 测试断言 mail/todo 拿到不同凭据）。另按评审要求新增**离线 Keychain 碰撞回归测试**（模拟 macOS 共享槽位语义，重新引入按账号缓存名会响亮失败，单独成提交）。工具数保持 62，文档守卫无需改动；CHANGELOG 已记路由与 logout 行为变更；本地 873 过（7 个 Windows 预存环境失败），ruff 干净，共 8 个提交。


mpalermiti:
Thank you for the rework. I'll start with what I verified, because it bears on what comes next.

**All three blockers from #61 are fixed.** I had this reviewed twice. The second pass looked only at those three blockers and the upgrade path, and it checked them against the installed SDK source instead of the PR description.

- **Shared Keychain cache:** one cache name plus per-account records is the right shape. Multiple credentials stay coherent because `PersistedTokenCache.modify` takes a cross-process lock, reloads when the signal file's mtime changes, and merges its writes.
- **Data paths:** every one fails closed. We enumerated all nine places that select an account.
- **Routing:** it refuses instead of guessing. Through mcp 2.2.0, the tool name the router reads is the name that gets dispatched.

**But I'm not going to take in-process multi-account, in this shape or a revised one.** That is a decision about the feature and not about this PR's quality. I would rather tell you now than after a third round.

The reason is risk, and it has two parts.

**The binding logic is hard to get right.** This change turns "whose mailbox is this?" from a constant into a computed value. Three review passes each found new ways to attach the wrong identity to a name:

- Adoption copies the legacy login and doesn't move it, so editing `default_account` binds a second name with no login at all. `outlook-mcp status` is the command that performs the binding.
- `auth <name>` saves whoever completed the flow, without checking it against the name.
- Logout is undone on restart by the leftover legacy record.

Each of these can be fixed. What concerns me is that new ones kept turning up in every pass.

**And the risk doesn't stay in the auth module.** Anything that holds state across calls becomes a place where one account's state can be replayed against another.

- The pagination cursors are the example. Nobody designing the routing was thinking about them, and a cursor minted on one inbox was silently applied to another after a switch.
- Every future tool that keeps a cursor, a delta token or a cached client inherits that exposure.
- The failures are quiet ones: mail goes out from the wrong account while every surface reports the right one.

I'm not willing to carry that in a server that can send and delete mail.

**Your need is real, and I want to meet it a different way.** Your own examples route mail to one account and to-do to another. One server process per account does that with configuration alone.

- Each process knows exactly one identity, so a wrong identity is structurally impossible.
- `OUTLOOK_MCP_TOOLSETS` already ships. Run a mail-only instance on account A and a to-do-only instance on account B. That is 19 + 10 tools, against 62 for one full instance.
- The only missing piece is that the config directory is hardcoded. There is a working spike on `spike/config-dir-override`.

If your setup needs something this model can't do, such as reading both accounts in one call or switching mid-conversation, tell me. I'd rather hear it now.

**One design constraint: the override must move only the config directory, never `HOME`.**

- The MSAL signal file lives at `~/.IdentityService/`.
- If both processes share that file, they share the lock-and-merge path into the one Keychain item. That is the mechanism you proved correct.
- If each process gets its own `HOME`, each gets its own signal file and they clobber each other's token. That is #61's original bug.

So the new model is built on your keychain work. It doesn't throw that work away.

**The PR I'd love from you, if you want it:**

1. **Config override:** add `OUTLOOK_MCP_CONFIG_DIR`, starting from the spike, plus a README section showing two client entries with toolsets.
2. **Remove the broken scaffolding on `main`:** `switch_account`, `list_accounts`, `_credentials`, and the `accounts` and `default_account` config fields. Accept and warn on those fields so old configs still load.
3. **Keep the collision test:** carry over `tests/test_keychain_collision.py`. It now guards the two-process model.
4. **Two standalone fixes from this PR:**
   - The lifespan should catch `OSError` and `PermissionError`, not only `ValidationError`.
   - `main` has the same `except ValueError` branch that, as far as I can tell, azure-identity's `@wrap_exceptions` makes unreachable. It's worth a look.
5. **The live proof I can't run myself,** because you have two accounts and I don't. I need to see two processes, two identities and one Keychain item coexisting:
   - Auth account A in one config dir and account B in another.
   - Confirm both authenticate.
   - Confirm a token refresh on one doesn't evict the other.

If you'd rather not, no hard feelings. I'll build it myself and credit the keychain finding in the CHANGELOG.

**One thing for your own machine, regardless of the above.** `tests/test_cli.py:36` and `:47` delete the real `~/.outlook-mcp/auth_record.json`. `_patch_config` patches only `load_config` and never `_auth_record_path`. Don't run that file locally without a redirected `HOME`.

I have full notes from both passes, about twenty findings. Most are moot under this direction, but I'm happy to share them if they'd be useful to you.

The README has claimed multi-account support since before it worked. That is mine to fix, and it is part of why this PR got as far as it did. I'll close this once you've had a chance to read it.

Nyaecho:
Thank you for the double review — and for making the call now rather than after a third round. I accept the direction.

The cursor finding is what settles it for me as well. I guarded `outlook_changes_since` in the first cut, but "every future tool that keeps a cursor, a delta token or a cached client inherits that exposure" is a maintenance property, not a bug list — it would keep generating exactly the kind of findings each of your passes found. The two-process model removes the class instead of enumerating its members.

It also covers my real deployment as-is: mail/calendar/contacts on one account, To Do on the other — your 19+10 example is literally my config, and `OUTLOOK_MCP_TOOLSETS` already does the splitting. So close away.

I'll take the PR you sketched:

1. `OUTLOOK_MCP_CONFIG_DIR`, starting from `spike/config-dir-override`, plus the README section showing two client entries with disjoint toolsets — and the constraint stated as you put it: the override moves the config directory only, never `HOME`, so both processes share the one MSAL signal file and merge into the one cache entry.
2. Remove the `switch_account` / `list_accounts` / `_credentials` scaffolding and the `accounts` / `default_account` config fields, accept-and-warn on legacy shapes.
3. Carry `tests/test_keychain_collision.py` over, reworked to assert the two-process invariants: one cache name, the shared signal file, records pinning identities.
4. The two standalone fixes — lifespan catching `OSError`/`PermissionError`, and the `except ValueError` branch on main (I'll verify it against azure-identity's `@wrap_exceptions` and fix or document what I find).
5. The live proof, since I do have two accounts: two config dirs, two processes, one Keychain item — auth both, refresh one, confirm the other survives. I'll post the transcript here.

And yes, please share the ~20 findings — the ones that survive the direction change, especially anything on the lifespan/error-handling paths item 4 touches, would save me a pass.

## 内部备注（不贴出）
- 他读后会自行关闭 #66（"I'll close this once you've had a chance to read it"），我们无需请求保持开启。
- 新 PR 完成清单见此；spike 基 c7de41b 已确认存在。

mpalermiti:
Closing as agreed — thank you for taking it the way you did. Here are the findings from both passes that survive the direction change, grouped by what they touch. Line numbers are against `190f313`. Most of the routing/gate items are moot now and omitted.

**Lifespan and error handling (item 4 of the narrowed PR)**

- `server.py:61` — the lifespan guard catches only `ValidationError`. `load_config()` also raises `PermissionError` (symlinked config, `config.py:288`) and `OSError` from the `chmod` at `:292`; both escape and kill the stdio server with the remedy on stderr. `cli.py:44` already catches the wider set.
- `auth.py:413` — the `except ValueError` / `_is_azure_unencrypted_refusal` branch and the `except AzureError` "transient" branch are unreachable for anything raised by `get_token`: azure-identity decorates `_acquire_token_silent` and `_request_token` with `@wrap_exceptions` (`_internal/decorators.py:63-82`), which retypes every non-`ClientAuthenticationError` into `ClientAuthenticationError`. Concrete cost on Linux with PyGObject importable but no Secret Service: the `ValueError("...allow_unencrypted_storage...")` is converted before your `isinstance` check, so `UnencryptedTokenCacheError` never raises and every tool says "re-run auth" forever instead of naming the config fix. `main` has the same shape. `test_transient_token_error_does_not_flip_anything` only passes because a MagicMock bypasses the wrapping.
- `auth.py:558`, `:473`, `server.py:1043` — refusals are bare `ValueError` → `ToolInputError(code="invalid_input", action=None)`. That tells the model its *arguments* were wrong when it's server policy. For the config-dir model this mostly goes away, but the pattern is worth not carrying over.

**Config validation**

- `config.py:214` — a misspelled capability key warned and silently rerouted; `config.py`'s default `extra="ignore"` drops a misspelled top-level key with nothing at all. Moot for routing, but `extra="forbid"` (or explicit accept-and-warn per field) is worth keeping in mind for the legacy `accounts`/`default_account` handling.
- `config.py:61` — `_check_account_name` rejected Windows-reserved words on every platform and >64 chars, bricking installs 1.21 accepted. Moot once the fields go, but the accept-and-warn you're planning for legacy shapes should be the whole story.
- `auth.py:202` — `_account_config` never fell back to top-level `tenant_id` despite its docstring; and azure-identity pops a passed `client_id` when a record is present (`interactive.py:102`), so a configured `client_id` is a no-op on refresh paths while `tenant_id` is honoured (`:103`). Worth a comment wherever the single-account path builds a credential from a record.

**Keychain / cache — for the reworked collision test**

- The Keychain item is shared host-wide across *every* azure-identity persistent cache of any name (`_persistent_cache.py:64,95` hardcode service `Microsoft.Developer.IdentityService`, account `MSALCache`); only the signal and lock files are per name. A foreign azure-identity app's write is invisible to outlook-mcp's instances and gets overwritten on the next `modify`. Pre-existing, fail-closed (re-auth, not a leak), but the comment at `auth.py:49-58` overstates what one name buys. Your two-process test should assert the shared signal file explicitly — that's the invariant that makes the merge path work.
- CAE: `graph.py:24` sets `is_cae_enabled=False`, and kiota passes it on both branches. A CAE cache would be a second `PersistedTokenCache` on the same Keychain item with a *different* signal file — a real clobber. Worth a one-line comment there saying it's load-bearing.
- `login_interactive`'s docstring (`auth.py:274-276`) claims it completes silently from cache; it can't — no record is passed, so `interactive.py:290,311` raises and it always prompts. Safer than documented; fix the docstring.
- `auth.py:310` reads the private `cred._auth_record`. If that attribute goes, auth prints success and writes nothing.

**Tests**

- `tests/test_cli.py:36,47` — repeated because it matters on your machine too: `cmd_logout` reaches the real `~/.outlook-mcp/auth_record*.json`. Patch `auth._auth_record_path` to `tmp_path`.
- `tests/test_routing.py:52` checked the test's own constant rather than `routing._GROUP_CAPABILITY`. Moot with routing gone.
- `_save_auth_record` writes non-atomically with a 0644 window while `config._atomic_write` exists.
- `assert self._active_account is not None` was control flow — stripped under `python -O`.

**Pagination — the one that generalises**

- `pagination.py:90` — cursors and delta tokens carried no scope identity. For the two-process model this is structurally solved, but `calendar_read.py:316` already pins `{"calendar": id}` into the cursor for the same reason within one account. Any future scope dimension wants the same treatment; worth a sentence in `pagination.py`'s docstring.

Also for the record: the README has claimed multi-account since before it worked — that's on me and the config-dir PR is where it gets made honest. The keychain finding is yours and will be credited in the CHANGELOG.

PR #67, opened 2026-09-17
feat(todo): task detail with checklist items and task attachments

Nyaecho:
Eight new To Do tools, taking the surface from 62 to 70:

- `outlook_get_task` — one task's full detail: `checklist_items` (ordered unchecked-first, matching the To Do client), notes, due, recurrence flag, attachments count.
- `outlook_add_checklist_item` / `outlook_update_checklist_item` / `outlook_delete_checklist_item` — add a sub-step, rename it or check it off, remove it.
- `outlook_list_task_attachments` / `outlook_download_task_attachment` / `outlook_upload_task_attachment` / `outlook_delete_task_attachment` — attachment listing; download confined to the configured attachments directory (the same rule as mail attachments); upload via upload session (0–25 MB); delete.

Includes a fix found in production use: the first cut of `outlook_get_task` crashed on any task that had sub-steps — the SDK's `TodoTask.checklist_items` is a plain list (there is no `.value`), and the unit-test mock had mirrored a `.value` shape the wire never produces. The mock now mirrors the real shape, and the fix was verified read-only against a live mailbox (303/303 tasks, checked-item ordering correct).

Bookkeeping: tool count 62 → 70 everywhere it's asserted or narrated — README (category list, toolsets section, examples), SKILL.md, server.json, the test guards; the ROADMAP "Checklists" line is removed as shipped. Tool-surface budget ceiling 12,000 → 13,500 (measured 12,712 for the 70-tool surface). CHANGELOG entry under Unreleased. Three commits; suite 845 passed locally (7 pre-existing Windows-only environment failures, unrelated); `ruff check` clean.

Independent of the multi-account rework — it touches only the To Do toolset, docs, and guards, so the two can land in either order.

## 中文说明

八个新 To Do 工具，工具面 62 → 70：`outlook_get_task`（单任务完整详情，含 `checklist_items` 子步骤——未勾选在前，与 To Do 客户端一致——及备注/截止/重复标志）；子步骤增改删三件套（`outlook_add_checklist_item` / `outlook_update_checklist_item` / `outlook_delete_checklist_item`）；任务附件四件套（列表/下载/上传/删除，下载限制在配置的附件目录内与邮件附件同规则，上传走 upload session 0–25 MB）。含一个生产环境发现的修复：`outlook_get_task` 初版对任何带子步骤的任务都会崩溃——SDK 的 `TodoTask.checklist_items` 是原生 list（无 `.value`），单测 mock 曾仿造 wire 上不存在的 `.value` 形状，现已按真实形状 mock，并对真实邮箱只读验证 303/303 任务全过、勾选排序正确。账目已对齐：README/SKILL/server.json/测试守卫的计数 62→70，ROADMAP "Checklists" 条目按已实现移除，工具面预算天花板 12,000→13,500（实测 12,712），CHANGELOG 有条目。与多账号 PR 相互独立，可任意顺序合入。三个提交；本地 845 过（7 个 Windows 预存环境失败），ruff 干净。


mpalermiti (changes requested):
Good feature and the hard part is right — both new filesystem paths go through `resolve_attachment_path` and rebind the resolved value, which is the one thing in this PR I'd have held it for. Every SDK field and builder name checks out too, including the `attachment_info` vs `attachment_item` trap the mail path would have invited.

Requesting changes on the upload/download pair, a permission that widened without the docs following, and two release-process rows. Static review — I ran no tests against my mailbox.

## Blocking: the upload and download path

**1. `upload_task_attachment` reports success whether or not the upload worked** (`todo_attachments.py:154`). `_upload_large_file` discards every chunk PUT's response (`mail_attachments.py:169`), and httpx doesn't raise on 4xx/5xx — so a 401 on an expired session token, a 413, a 429 or a 507 advances `offset` and exits the loop normally. The tool returns `{"status":"uploaded","name":…,"size":…}`, the agent tells the user the file is attached, and `outlook_list_task_attachments` then returns count 0.

Mail only reached this above its 3 MB inline threshold. Your own module docstring says every size goes through `createUploadSession` + chunked PUT, so here the window is every byte. Every other raw-httpx path in the repo calls `raise_for_status()` — `_delta.py:206`, `mail_read.py:605`. The one unit test monkeypatches the helper away (`tests/test_todo_attachments.py:162`), so CI can't see it.

**2. `download_task_attachment` truncates the destination before it has any bytes** (`todo_attachments.py:92`). `open(save_path, "wb")` empties an existing file, and `.content.get()` is typed `Optional[bytes]` — kiota returns None for an empty body. So a 0-byte `taskFileAttachment`, which the To Do web and mobile clients can create and your own 0-byte guard only covers uploads *you* perform, leaves a zeroed file and raises `TypeError: a bytes-like object is required`. `_wrap_tool_errors` doesn't recognise TypeError as a Graph error, so the model gets a bare crash with the message withheld.

The part that makes this more than a bug: the zeroed file is inside `attachments_dir`, which is also where outbound files are staged — so a later `outlook_send_with_attachments` can send the emptied file under its trusted name. Write to a temp path and move it into place, or check the body first.

**3. An upload can hang forever** (`mail_attachments.py:161`). The loop advances by `len(chunk)`. `os.path.getsize()` (`todo_attachments.py:120`) and the read loop are separate filesystem operations with an `await` between them, so if the file shrinks in that window — a concurrent turn rewriting a staging file, a sync client re-materialising it, a network mount — `f.read()` returns `b""` while `offset < file_size`. Then `chunk_size` is 0, the `Content-Range` is nonsense, `offset` never advances, and it PUTs empty chunks indefinitely inside a tool call. `if not chunk: break`.

**4. No timeout, and no retry** (`mail_attachments.py:158`). `httpx.AsyncClient()` with no `timeout=`, bypassing `throttle.send_with_retry`. httpx's default is 5 seconds, and `_UPLOAD_CHUNK_SIZE` is 3.2 MB — so a healthy connection on a modest uplink raises `WriteTimeout` mid-upload, and a 25 MB To Do attachment is eight consecutive exposures. On 429 there's no Retry-After handling at all, since kiota's RetryHandler is bypassed too. `_delta.py:201` and `mail_read.py:593` are the two paths `throttle.py`'s docstring says must self-retry; this makes the uploader a third without adding it to that contract.

## Blocking: a permission got more powerful and the docs didn't follow

**5.** `README.md:489`'s risk table still reads `todo_write` — "Safe — your own task list". After this PR that one grant also reads a local file and pushes its bytes to Graph (`outlook_upload_task_attachment`, gated only on `CATEGORY_TODO_WRITE` at `todo_attachments.py:113`) and irreversibly deletes attachments. The mail-side equivalent requires `CATEGORY_MAIL_SEND`.

Three lines below that table is my own "draft-only assistant" example — `{"read_only": false, "allow_categories": ["mail_drafts","mail_triage","todo_write"]}` — which anyone copying now grants more than they think. The PR touches README lines 24/44/59/68/338/431/550 but not this table, and the confinement callout at `README:366` still names only the three mail tools. Either update both, or put upload/delete behind their own category.

## Blocking: two release rows

**6. `scripts/preflight.py` has no new `ENDPOINTS` rows.** `RELEASING.md:17` says it outright: add a row when a tool hits a Graph endpoint family not yet covered. `ENDPOINTS` still holds only `("me/todo/lists", "To Do")` — nothing probes `checklistItems`, `attachments`, `$value` or `createUploadSession`. `classify()` treats 403/501 as a release blocker, so a family Graph refuses on consumer mailboxes ships green to PyPI, ClawHub and the registry. That is verbatim what happened to the four `/me/mailboxSettings/*` tools in v1.7.0, yanked in v1.7.1 — and `taskFileAttachment` support on consumer outlook.com is exactly the unverified assumption here. I know these rows need a live task id, so it isn't the one-liner the existing entries are; say so if you'd rather I take it.

**7. `ci.yml:107` still expects 62 tools** while `ci.yml:60` was swept to 70. Correct today, because PyPI serves 1.22.0 — and permanently red from the moment 1.23.0 publishes. It fails at the count assert, *before* the "CLI entrypoint must be wired and runnable" step, so the weekly canary stops verifying anything about the published package, and a permanently-red scheduled job is the one that gets muted. `RELEASING.md` §3's lockstep list doesn't name `ci.yml` and no test reads it, so nothing else will catch this.

## Not blocking

- **`_resolve_list_id` (`todo.py:59`) gates on truthiness.** `list_id=""` — a very common LLM/JSON-schema outcome for an optional string — is falsy, so `outlook_delete_task_attachment` and `outlook_delete_checklist_item` silently retarget the default list instead of raising. It's also the only Graph id in the To Do surface never passed through `validate_graph_id` (`task_id`, `attachment_id` and `checklist_item_id` all are), and `me.todo.lists.get()` at line 62 sends no `$top` and ignores `odata_next_link`, so a user whose first page lacks the `defaultList` entry gets an arbitrary list.
- **`list_task_attachments` (`todo_attachments.py:43`) doesn't paginate** — bare `.get()`, no `$top`, `odata_next_link` discarded — and unlike every other list tool emits no `has_more`/`next_cursor`, so a caller can't detect the truncation while `count` looks authoritative. `list_tasks` (`todo.py:217-237`) is the shape to copy. Same class as the `childFolders` bug we already ate once.
- **`str()` on a kiota datetime isn't ISO** (`todo.py:133`, `:140`, `todo_attachments.py:57`). `checked_date_time` deserializes to a `datetime`, so you get `"2026-09-15 09:00:00+00:00"` — space, no `T`, no `Z` — in the same response as `due`, which is Graph's raw ISO string. One result, two incompatible formats. The tests can't see it because the mocks feed strings and assert the string back.
- **`add_checklist_item` (`todo.py:496`) dereferences `response.id` unguarded.** `post()` is `Optional[ChecklistItem]`; an empty 201 or a 204 raises `AttributeError` *after* the sub-step exists, the agent retries, and the task now has two. `mail_attachments.attach_to_draft:352` guards the identical pattern. Same shape at `todo_attachments.py:154` and `todo.py:267`.
- **Three new payload builders aren't in the wire guard.** `TestTodo` in `test_write_payloads_reach_the_wire.py` still has only create/update task. `add_checklist_item`, `update_checklist_item` and `upload_task_attachment` are never serialized through the backing-store proxy — their only payload checks read attributes back off the model just built, which is the exact shape that file exists because it's blind to. `update_checklist_item` is the one I'd want covered most: a partial PATCH where the caller picks the fields, i.e. #63's leak shape. (No `None` actually leaks today — I checked. This is a missing guard, not a live bug.)
- **`test_live_todo_write.py:157` asserts an order Graph doesn't promise.** `todo.py:147` says expansion order is undefined; both items are unchecked and the sort is stable, so this rides on Graph's ordering and can fail correct code at the pre-tag gate. Line 172's `!= []` is vacuous — the `next(...)` below would already raise.
- **`test_live_query_shape.py:255` is tautological.** It asserts the client-side sort `_checked_last` just applied, and `isinstance(..., list)` holds for `[]` — so it runs happily against a task with no sub-steps, which is the condition that let the 1.22.0 crash ship. It needs a `checklist_count > 0` precondition or the write tier should own the guarantee.
- **`conftest.py:69`'s live-skip was loosened on a justification `auth.py` contradicts.** The comment says AuthManager handles a config with per-account `client_id`; `try_cached_token()` short-circuits on exactly that shape (`auth.py:243`), `_make_credential` passes None through, and `load_config` never promotes it. So the tier doesn't become runnable — it skips one fixture later with the wrong remedy. It's also an unrelated change to shared test infrastructure inside a PR about To Do tools, and nothing covers it. I'd pull it out.
- Smaller: the module docstring says there's no inline base64 POST for this resource, but the installed SDK has `attachments_request_builder.post(body: AttachmentBase)` and `TaskFileAttachment.content_bytes` — and that claim is what forces a 2 KB file through a session + separate PUT; `outlook_download_task_attachment` and `outlook_update_checklist_item` are missing from `toolsets.IDEMPOTENT` although their twins are in it; `raise FileNotFoundError` (`todo_attachments.py:118`) isn't a `ValueError`, so `_wrap_tool_errors` re-raises and the path is withheld from the model; "0–25 MB" is advertised in six places while a 0-byte file is hard-rejected; `content_type` isn't sanitized on the listing; `update_checklist_item` validates after `_resolve_list_id`'s network call while `add_checklist_item` validates before; and `_resolve_list_id` fires an uncached `GET /me/todo/lists` on all 13 tools, roughly doubling the requests in a normal checklist flow.
- The `~8.6k` token baseline is now ~13% low in `README:442`, `toolsets.py:10` and `CLAUDE.md:82` — this PR measured 12,712. Worth correcting while you're in there.

## Checked and clear, so nobody re-spends it

Both new filesystem paths call `resolve_attachment_path` before any I/O and rebind the resolved value. No `None` is assigned to any SDK model field. `$expand` reaches the wire as `["checklistItems"]`. `validate_graph_id`'s charset doesn't reject real Graph ids. All five write tools call `check_permission`. The 62→70 sweep and the `TOOL_SURFACE_CEILING` raise are internally consistent. `_PAIRS` in `test_select_covers_the_formatter.py` correctly doesn't apply — there's no `$select` anywhere in `todo.py`.

The four upload/download items are the ones I'd not ship. The rest is mostly small.

mpalermiti:
Correcting my own review: I framed finding 1 as "uploads can fail and still report success." I now think it's worse than that — **I don't believe a To Do upload can succeed at all**, and the reason the PR looks green is that the live test which would have caught it cannot run.

## The upload is unauthenticated

`_upload_large_file` sends exactly two headers (`mail_attachments.py:165-168`):

```python
headers = {
    "Content-Range": f"bytes {offset}-{end}/{file_size}",
    "Content-Length": str(chunk_size),
}
```

No `Authorization`, no `Content-Type`. That is correct **for mail**, because `createUploadSession` on a message returns a pre-authenticated `outlook.office.com` URL — an opaque external endpoint that carries its own token in the query string.

To Do's session is not that. The upload target is a Graph route, and the SDK's own generated surface shows the asymmetry cleanly:

```
attachment_sessions builders under users/item/messages/ : 0
attachment_sessions builders under users/item/todo/     : 8
```

`.../todo/lists/{id}/tasks/{id}/attachmentSessions/{sessionId}/content` exists as a generated builder because it is a `graph.microsoft.com` path — and Graph paths require `Authorization: Bearer`. Microsoft's own "attach files to a To Do task" flow sends the bearer on each PUT. So every chunk should come back `401 InvalidAuthenticationToken`, and because nothing checks the response (my finding 1), the tool returns `{"status":"uploaded"}` and no attachment exists.

The repo already has the right pattern twice — `_delta.py:59` and `mail_read.py:584` both do `credential.get_token(GRAPH_TOKEN_SCOPE)` and set the header. The awkward part is that `_upload_large_file` takes no credential and `upload_task_attachment` only receives `graph_client`, so this needs a signature change rather than a one-liner.

**Flagging my confidence honestly:** this is a structural inference from the SDK layout and Microsoft's documented flow, not something I've observed. It needs one live upload to settle — which brings me to why nobody has seen it.

## The live test that would have caught it cannot run

`test_live_todo_write.py:198`:

```python
base = Path(live_write_config.attachments_dir)
```

No `expanduser`. `attachments_dir` defaults to the literal string `"~/.outlook-mcp/attachments"` (`config.py:39`), and `SKILL.md` tells users to put exactly that in their config. So:

```
test writes to   : ~/.outlook-mcp/attachments/…-src.bin   (is_absolute: False)
tool resolves to : /Users/<you>/.outlook-mcp/attachments/…-src.bin
```

`resolve_attachment_path` *does* expanduser (`mail_attachments.py:45/53`). So the test creates a directory literally named `~` under pytest's CWD, drops 100 KB in it, and the tool then fails `FileNotFoundError` before a byte reaches Graph. `TestAttachmentRoundTrip` — the tier whose docstring says it covers "the one thing mocks cannot vouch for" — can never go green, and it leaves a stray `./~/` tree in the repo either way.

The teardown is asymmetric in the same direction: had the upload worked, `download_task_attachment` would write into the real home directory while `dst.unlink(missing_ok=True)` (line 243) deletes the *relative* path, silently leaving the real file behind.

Root cause worth fixing once: `live_write_config` (`conftest.py:106`) hands back the user's real config unmodified, and this is the first live test to touch the filesystem. It should sandbox `attachments_dir` to a tmp_path rather than have each test remember to expand it.

## Three more from the same sweep

**Orphaned upload sessions on a live task.** `todo_attachments.py:149-154` POSTs `createUploadSession` and then calls the uploader with no `try`/`except`. Any failure after the POST — the 401 above, a timeout, the shrunk-file loop — leaves an open session attached to the task until Graph expires it, and every "retry for safety" mints another. `AttachmentSessionItemRequestBuilder.delete()` exists, so the cancel is one call. There's no `attachment_sessions` reference anywhere in `src/` today. Mail has the same shape but its sessions hang off a draft the user can delete; these hang off a live task.

**The multi-chunk path has zero coverage at every tier.** The live payload is 102,400 bytes against `_UPLOAD_CHUNK_SIZE` of 3.2 MB — 32× smaller than one chunk — so `while offset < file_size` runs exactly once. Nothing anywhere exercises offset advancement, the `Content-Range` arithmetic across chunks, or the empty-read hang. Related and worth fixing together: the SDK documents that Outlook attachment sessions return `next_expected_ranges` as a single `{start}` telling you where the next upload must begin, and `_upload_large_file` ignores it entirely, blindly advancing by the local chunk length — so a partially-accepted chunk desynchronises and corrupts the attachment. A 4 MB live payload exercises two chunks and stays well under the 25 MB cap.

**`_checked_last` doesn't deliver what it promises.** `todo.py:153` sorts on the boolean alone, and Python's sort is stable — so order *within* the unchecked group is whatever Graph returned, which the function's own docstring says is unspecified. The docstring also says "the order we emit is load-bearing, not cosmetic," and `outlook_get_task` promises items "ordered unchecked-first, matching the To Do client." Two identical calls on an unchanged task can return a different first unchecked item, so an agent reading "the next step" gets a different answer each poll and a diff-based agent sees a phantom change. `_format_checklist_item` already emits `created` (`todo.py:140`), so the fix is sitting there: `key=lambda i: (i["is_checked"], i["created"])`.

---

Revised ordering, if it helps: the auth header and the live-test path are the pair that matter — together they mean `outlook_upload_task_attachment` ships broken *and* untested. Everything else in my earlier review stands.

Still static analysis; I ran nothing against my mailbox.

Nyaecho:
Thanks for the exceptionally thorough review — and for the follow-up, which was right in a way even it hedged on. I ran the live experiments against a consumer outlook.com mailbox (NyaechoNeko@outlook.com), and they settle all four upload/download items at once:

| Experiment | Result |
|---|---|
| `POST .../tasks/{id}/attachmentSessions` | **404** (list/task creation on the same account 201s, so not a path error) |
| Inline `POST .../tasks/{id}/attachments` with `{@odata.type: taskFileAttachment, name, contentBytes, size}` | **201** at 64 B, 512 KB, 3 MB, and 4 MB |
| 25 MB inline POST | **400** `request body size limit 30000000` — a hard JSON cap; ×4/3 base64 puts the honest ceiling near 22.5 MB raw |
| `GET .../attachments/{id}/$value` + Bearer | **200**, bytes verified; the entity GET also returns `contentBytes` |

So the follow-up was correct with the caveat removed: on consumer accounts the session path isn't "can fail silently" — it's *unreachable*, and the missing `Authorization` never even got a chance to matter. **The session/chunked-PUT path is gone entirely.** Upload is now a single inline base64 POST of a `TaskFileAttachment` through the SDK (`attachments_request_builder.post`), so kiota's throttling/retry applies and no raw-httpx path remains to hang, time out, or mis-report. That one rewrite dissolves findings 1, 3, 4, the follow-up's missing-auth finding, and the orphaned-session concern (there is no session to orphan). The client ceiling is **1 byte – 20 MiB**, checked with `os.stat` before any I/O — the 30 MB body cap × 4/3 base64 leaves no honest room above that, and the error message says so. A null response id raises with "the attachment may exist — read back, don't retry" rather than an AttributeError.

**Finding 2 (download truncation):** fixed the way you offered — all bytes are fetched *first* (entity GET + `contentBytes`, matching `mail_attachments.download_attachment`'s style; kiota decodes the base64), then written via temp file + `os.replace`. A 0-byte attachment writes an honest 0-byte file, and a failed fetch never touches the staged file (there's a unit test that stages bytes, fails the fetch, and asserts the file survives).

**Finding 5 (docs):** took the update-both route — the risk table's `todo_write` row now says what the grant actually carries (reads local files from `attachments_dir`, uploads their bytes, irreversible deletes), the draft-only example gained a pointer, and the confinement callout names the To Do attachment tools behind the same fence. No new category.

**Finding 6 (preflight):** added `_todo_task_rows` — two GETs provision a live task, then it probes `checklistItems`, `attachments`, and `$value` (the last only when an attachment actually exists; otherwise a visible SKIP). The inline POST itself stays out: the script's contract is read-only, same caveat as the `$batch` note already in the file, and the GETs surface the same 403/501 not-supported signal. If you'd rather own a POST-row refactor, happy to leave it to you — the table above is the evidence either way.

**Finding 7 (ci.yml):** left at 62 (that's what PyPI serves today) with a "bump at release time" comment at the assert, and `ci.yml` is now on RELEASING.md §3's lockstep list.

**Not-blocking — all addressed:** `_resolve_list_id` rejects `list_id=""` explicitly, validates through `validate_graph_id`, pages the default-list lookup (`$top` + nextLink), and caches the resolution per `GraphServiceClient` in a `WeakKeyDictionary`; `list_task_attachments` paginates with `has_more`/`next_cursor` in `list_tasks`' shape; kiota datetimes go through an `_iso_datetime` helper (datetime → `isoformat()`, strings untouched) across the To Do surface; `add_checklist_item`, `create_task`, `get_task`, and the attachment POST all guard null responses with "read back, don't retry"; the three payload builders are in the wire guard through the backing-store proxy (update's partial PATCH asserts `isChecked`/`checkedDateTime` stay *absent* — #63's leak shape, which your merge of #63 now also guards on the contacts side); `outlook_download_task_attachment` and `outlook_update_checklist_item` joined `IDEMPOTENT`; the missing-file error is a `ValueError` carrying the path so it reaches the model; "0–25 MB" is gone everywhere in favor of "1 byte – 20 MiB"; listing `name`/`content_type` pass `sanitize_output`; `update_checklist_item` validates before the network call; `_checked_last` sorts on `(is_checked, created)`; the module docstring no longer claims inline POST doesn't exist — it now records the 404 evidence instead; the conftest live-skip relaxation is reverted; and the ~8.6k baseline is corrected to the measured 70-tool number (12,855, ceiling re-anchored at 13,650).

**Live-test sandboxing:** `live_write_config` now hands back a copy with `attachments_dir` pointed at a per-test `tmp_path` — the root fix for the `~`-literal bug you found, so `TestAttachmentRoundTrip` can actually go green (512 KiB binary payload through the inline path) instead of leaving a `./~/` tree behind.

One note back at you: `mail_attachments.py` raises the same `FileNotFoundError` you flagged here (lines 206/333) — same `_wrap_tool_errors` consequence, out of this PR's scope, flagged for a follow-up.

Rebased onto current `main` (6c70b55, post-#63/#68) before pushing; CI green. Suite: 925 passed / 7 Windows-environment failures (unchanged baseline) / 46 deselected; ruff clean. Seven new commits on top of the original three.

mpalermiti:
Checking in — the one thing between this and merge is the upload-size question from the re-review: the probe hit `POST …/attachmentSessions`, which the SDK only exposes as GET, so a 404 there is expected on any account and doesn't tell us whether sessions work on consumer mailboxes. Session creation is `POST …/attachments/createUploadSession`.

Two live calls settle it: one `createUploadSession` on the correct URL, and one inline POST near the 20 MiB ceiling. If either is a hassle, capping at the 4 MB you've actually verified is a fine answer too — it can always be raised later with evidence.

Everything else you did checked out. No rush; just didn't want it to look like it had gone quiet on my side.

Nyaecho:
You were right on both counts — the URL my probe hit, and what the real answer would look like. Corrected numbers, all from the same consumer mailbox (NyaechoNeko@outlook.com):

| Call | Result |
|---|---|
| `POST …/attachments/createUploadSession` (body keyed `AttachmentInfo`) | **201** — sessions do exist on consumer mailboxes; my earlier 404 was the wrong route (`…/attachmentSessions`), exactly as you said |
| session `uploadUrl` | a `graph.microsoft.com` route, `nextExpectedRanges: ["0-"]` |
| first-chunk PUT without `Authorization` | **401** `InvalidAuthenticationToken: Access token is empty` — your follow-up's prediction, verified verbatim |
| first-chunk PUT with `Authorization`, no `Content-Type` | **400** "Write requests (excluding DELETE) must contain the Content-Type header declaration" — a second header the mail-style uploader never sends |
| inline POST at exactly 20 MiB | **201** (first attempt hit a transient connection drop; clean 201 on retry) |
| inline POST at 25 MB (earlier probe) | **400** `request body size limit 30000000` |

So the session route is real — but every chunk needs `Authorization` *and* `Content-Type`, the response needs checking, and the offsets should follow `nextExpectedRanges`. Rather than grow this PR into a second, riskier upload path, I've kept it inline-only at **1 byte – 20 MiB**, now verified at both ends of the range, and corrected the module docstring and CHANGELOG — they had recorded my wrong-URL 404 as evidence that sessions don't exist on consumer accounts, which it isn't. Sessions are documented as the future route for >20 MiB, with the three requirements above named, if anyone actually needs it.

Everything else unchanged: 925 passed / 7 Windows-environment failures / ruff clean on top of current main. The only commit since the last green CI is the docs correction above — GitHub hasn't folded it into a run as I write.

## 内部备注（不贴出）
- 发帖前动作：在分支上补一个小提交，把 todo_attachments.py 模块 docstring 与 CHANGELOG 里"session 404 on consumer"的断言改为修正后事实（session 存在，需 Bearer+Content-Type+响应检查，inline ≤20MiB 是本 PR 的选择），推送过 CI 后再贴本回复。
- 修正探针脚本存档 `pr-replies/probe2-corrected.py`。

mpalermiti:
Thanks for running the corrected probes — that table is exactly the evidence the decision needed, and the 401/400 pair on the chunk PUT settles the follow-up's inference either way. Inline-only at 1 byte – 20 MiB, verified at both ends, is the right shape for this PR; the session route with its three requirements is well recorded for whoever needs >20 MiB.

One mechanical thing before CI can run: #76 landed on `main` this morning and this PR now conflicts — `tests/test_tool_surface_budget.py` is the file. Resolve the ceiling to your 13,650 (#76 added +143 to the surface; your "~6% headroom" note becomes ~4.8%). Rebase and push and I'll do the last pass on the rebased head.

Separately: your last two comments carry a block headed 内部备注（不贴出）— that reads as private notes your tooling posted by accident. Nothing sensitive in them, flagging so you can edit them out if you'd like.

mpalermiti:
While you rebase — here's the rest of the re-review from the 18th, which I owed you and hadn't posted. It ran against `6ad4832`; your docs commit didn't move these lines. Every fix you listed in that round checked out as real, and no crash-class bug is left. These are the remaining items, roughly in order of what I'd want before merge.

**Would hold for**

1. **`todo.py:112` — the default-list page walk is an uncapped `while True` that can re-request page 1 forever.** `wrap_nextlink` turns a `$skiptoken` link into a cursor and `apply_pagination` sets `$skiptoken`, but `build_request_config` silently drops params the SDK query class lacks, and `ListsRequestBuilderGetQueryParameters` has no `skiptoken` — so every iteration sends `{top: 100}`. Reproduced offline. Trigger needs >100 lists plus a skiptoken link, so treat it as cheap insurance: follow the link with `with_url(next_link)` as `calendar_resolver.fetch_all_calendars` does, and cap the pages. `test_default_list_found_past_the_first_page` passes with `apply_pagination` no-op'd, so cursor forwarding is untested.
2. **`todo_attachments.py:131` — `attachment.content_bytes or b""` treats a missing `contentBytes` as an empty file** and replaces the staged destination with 0 bytes, returning size 0 as success — contradicting the docstring's "an empty or failed download never truncates a staged file". `attachment.size` is available to cross-check. If the response lacks `@odata.type`, kiota builds `AttachmentBase`, which has no `content_bytes`, and you get an opaque `AttributeError` (reproduced with the real parser).
3. **`todo_attachments.py:208` — the inline POST inherits kiota's default retry**, which re-sends the ~28 MB body on 429/503/504. Offline probe: 4 POSTs, 112 MB. A 504 *after* Graph committed creates duplicates silently, and the "read back, don't retry" guard can't see it. Pass `RetryHandlerOption(should_retry=False)` on this POST, and map transport timeouts to the existing `ValueError` so the agent gets the verify-first text rather than a bare crash.
4. **`README.md:493` — the risk-table row says `todo_write` gates task-attachment download, but `download_task_attachment` has no `check_permission`.** It matches the ungated mail twin, so the code is consistent; the README overstates the gate. Drop "download" from the row or gate both downloads.
5. **`CHANGELOG.md:36` — "Added: 8 new tools" is the only entry, but five shipped tools changed behaviour** through the shared `_resolve_list_id` and `_format_task`: `list_id=""` is now refused (on 1.22.0 it fell through to the default list); every explicit `list_id` runs through the Graph-id regex; `created` is now ISO with a `T`; the default list resolves once per process. All defensible, none has a "Changed" line, and a client that fills optional strings with `""` breaks on upgrade with no note.

**Worth fixing, not blocking**

6. `todo_attachments.py:177` — the size gate is `os.stat`, then an awaited lookup, then an unbounded `f.read()`. A file still being written bypasses the ceiling and `size` is stale (offline: gate saw 1,024 bytes, 26 MB were sent). `f.read(_MAX + 1)`, re-check, take size from `len(content)`.
7. `todo_attachments.py:133` — the temp file goes in `dirname(save_path)`; when `save_path` resolves to `attachments_dir` itself (`.` passes `is_relative_to`), that's the *parent*, and `mkstemp` writes beside `config.json` before `os.replace` raises. Also `sub/x.pdf` with no `sub/` → `FileNotFoundError` after the full fetch, not a `ValueError`. Refuse a directory or missing-parent destination before the Graph call.
8. `todo.py:142` — the "first list" fallback is cached for the client's lifetime with no eviction. One lookup that finds no `defaultList` pins the first list returned — possibly a shared one — for every later create/upload/delete until restart. Don't cache the fallback, or evict on 404. And CLAUDE.md/README both say "no local caching", so this wants a one-line carve-out either way.
9. `todo.py:343` — `get_task` reads the `$expand=checklistItems` list as complete and never checks `checklistItems@odata.nextLink`. Whether To Do pages the expansion is unverified; one task with 25 steps settles it.
10. `todo.py:226` — `list_task_lists` is still one unpaged `.lists.get()` while the default-list lookup pages. Same `with_url` helper fixes both.
11. `scripts/preflight.py:116` — the new probe hits `/attachments/{id}/$value`, which no tool calls, and skips the entity GET download depends on; `lists?$top=1` often returns an empty list so every To Do row can SKIP and the run still prints "Safe to tag". Probe the entity, harvest from the default list, report "N reachable, M not probed".
12. `tests/test_live_todo_write.py:227` asserts Graph's `size` equals the raw byte length — Exchange-backed attachments often report a size with overhead, so this may fail on the first pre-tag live run. And `:174` plus `test_live_query_shape.py:368` assert on output `_checked_last` just sorted by that key — same tautology shape as before.
13. `CLAUDE.md:82` / `README.md:446` / `toolsets.py:10` — the new "~12.9k tokens for 70 tools" is the chars/4 proxy set beside the o200k "~8,644 for 62", which `test_tool_surface_budget.py` says not to compare. On one unit it's 11,442 → 12,829 (+12%). Stale "~8.6k" remains at `server.py:104` and `tests/test_idempotent_hints.py:101`.
14. `tests/test_write_payloads_reach_the_wire.py:537` — the `update_checklist_item` wire test never asserts `isChecked` on the serialized payload; the only positive check is on the model attribute. One `assert_on_wire(..., '"isChecked": true')` case.

**Budget, for the record:** 70 tools measure 12,829 here (you quote 12,855); the 8 new ones cost ~173 each against an existing average of 185. The raise to 13,650 is proportionate. With #76 and #73 now on `main` the merged surface is ~13,000, so resolve to your number.

Checked and clean, so you don't re-trace: body-size math holds (20 MiB + 255-char name = 27,962,430 bytes, under the cap; exactly 20 MiB accepted, one byte more refused); the `WeakKeyDictionary` cache is keyed on `sdk_client` everywhere and a real `GraphServiceClient` is hashable and weak-referenceable; preflight's `_todo_task_rows` is read-only; `_iso_datetime` and the `(is_checked, created)` sort are correct; all 8 tools driven through a real SDK client on a mock transport produced correct URLs, methods and bodies with no nulls in partial PATCHes; the rebase onto #64/#65/#68 integrates with `assert_on_wire`, `_PAIRS` (no `$select` here, so it doesn't apply) and `consumer_mailbox_only` correctly.

Nyaecho:
Rebased onto main (8fea0d1, includes #76). The one conflict was the budget test; the ceiling stays 13,650 with the combined history noted in the header — re-measured honestly it reads 13,154 proxy tokens, ~3.6% headroom (the ~4.8% estimate missed that #76 grew get/update_event docstrings too, not just the new `timezone` parameter, and that my last docs commit had already spent +178 of the headroom).

All 14 items, one line each:

1. The `_resolve_list_id` walk now follows the raw `@odata.nextLink` via `with_url` with a page cap — rebuilding typed query params dropped `$skiptoken` (the SDK class has no such field) and re-fetched page one forever. The old pagination test only fed side-effect pages, so I replaced it with one that asserts the raw link reaches `with_url`.
2. A download with missing `contentBytes` is now an error, cross-checked against the entity's own `size`; the no-`@odata.type` case (kiota builds `AttachmentBase`) is a clear ValueError instead of a bare AttributeError. A genuine 0-byte attachment still writes an empty file.
3. The inline POST carries `RetryHandlerOption(should_retry=False)` — no 4× re-send of a 28 MB body, no 504-after-commit duplicates — and a transport timeout maps to the existing verify-first ValueError.
4. README risk row no longer says `todo_write` gates downloads; code unchanged, matching the mail twins.
5. CHANGELOG has the Changed rows for the five existing To Do tools (empty-string `list_id` now rejected, Graph-id validation, per-process default-list resolution, ISO datetimes).
6. The upload size gate is the read itself (`f.read(MAX+1)`), so a file growing between stat and read can't bypass the cap; a lying `os.stat` is proven insufficient in a test.
7. Download targets are validated before any Graph call: the attachments dir itself and missing parent dirs are both refused as ValueError.
8. The first-list fallback is no longer cached (only a found `defaultList` is); CLAUDE.md and README's "no local caching" claims carry the one-line carve-out.
9. `get_task` defensively follows a `checklistItems@odata.nextLink` if one ever appears (same `with_url` walk, page-capped); documented as defensive, no live write experiment run.
10. `list_task_lists` paginates via the same helper and emits the standard `has_more`/`next_cursor` pair (always terminal — the walk is server-side-complete).
11. Preflight now harvests from the default list (not `lists?$top=1`), probes the entity GET with `contentBytes` verified in the body, keeps `$value` as an extra row, and the summary reads "N of M probed endpoints reachable, K NOT probed" — "Safe to tag" is withheld whenever anything went unprobed.
12. Live assertions de-tautologized: checklist order is checked against expectations derived from the test's own actions (and, in the query-shape tier, that two identical calls emit identical order); the attachment size check relaxed to `>=` with the byte-for-byte download comparison as the real fidelity check.
13. Every surface figure now names its yardstick (o200k vs chars/4), with the same-yardstick comparison spelled out (11.7k for 62 tools → 13.2k for 70, +12%); the stale ~8.6k leftovers in server.py and the idempotent-hints docstring are gone.
14. `isChecked: true` finally has a positive wire assertion.

Final numbers: 989 passed / 7 failed — the same 7 Windows symlink/permission environment failures as before the rebase, no regressions; ruff clean; budget test passing at 13,154 under the 13,650 ceiling.

mpalermiti:
All fourteen verified in the code here — 997 green on the pre-rebase base, and the with_url walk, `_verified_content_bytes`, `should_retry=False`, the bounded read and the rest are all exactly as described. Nothing further on content.

One more rebase, sorry: yours landed onto `8fea0d1` (#76), and #73 merged on top of it since. The remaining conflict is `tests/test_live_query_shape.py` — both PRs appended a live guard to the same spot, so keep both. Merge-base target is `08cea09`. Once CI runs green on that head, this merges.

Nyaecho:
Rebased onto `08cea09` — the one conflict was the live-guard spot in `tests/test_live_query_shape.py`, both guards kept. CI green on `07429b1`; locally 1,028 passed with the same 7 Windows environment failures as before the rebase.

